PR #1, opened 2023-06-21
Adds altruism factor to agent decision making and configuration

WillemHueffed:
Adds altrusim factor in the config file

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -595,7 +595,7 @@ class Agent:
                     break
         elif self.ethicalTheory == "draftPick":
             ranking = ethics.findDraftPickRanking(self)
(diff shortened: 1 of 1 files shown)


nkremerh:
Code review first pass:

Overall does the expected behavior. Some style changes to include:

1. Let's rename altruismFactor into something like greedFactor or selfishnessFactor. This will help disambiguate this from the altruisticValue used in the utility calculus.
2. Line 41 of ethics.py, avoid using single character names. Maybe call it neighborhood?
3. Place the TODO comment that got moved back to just above line 70
4. Lines 444-449 are handling a separate issue, pull those changes out and put them in a separate branch

PR #4, opened 2023-06-23
Adds multiple debug modes to the simulation

WillemHueffed:
Adds debug mode options to README
Modifies config file debug mode option from boolean to string

### DIFF
--- a/README
+++ b/README
@@ -205,9 +205,12 @@ agentVision: [int, int]
     Set the distance in the four cardinal directions an agent can see.
     Default: [1, 6]
 
-debugMode: bool
+debugMode: [string, ...]
     Set the debug printing mode.
-    Default: False
+    Options: "agent", "all", "cell", "disease", "environment", "ethics", "none",  "sugarscape"
+    Default: "none"
+    Note: Some options may cause no output in current impementation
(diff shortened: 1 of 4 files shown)


WillemHueffed:
1. Done 
2. Updated - config file is now an array where multiple debug modes can be entered as CSVs. Control structure conditions now use if...in

WillemHueffed:
Third tries the charm?

nkremerh:
Final pieces:

1. Add some validation to verifyConfiguration in sugarscape.py to handle a few cases:
    a. Both all and none being specified should throw an error message and abort the program.
    b. If all is specified with subcomponents (e.g. [all, agent, disease]), the list should be condensed to just [all].
    c. If a subcomponent is specified which does not exist, throw an error message and abort the program.
2. Set the default debugMode to ["all"] in config.json.
3. The default debugMode in sugarscape.py (line 755) should be ["none"] within a list, correct?

Otherwise, this works perfectly. Good work.

WillemHueffed:
1. I used sets to keep the code relatively clean (saves some weird O(n^2) wonkiness) -  hope that's okay
2./3. Thanks for catching those!

nkremerh (approved):

Issue #7, opened 2023-06-28
Adding new agent field seems to break determinism


WillemHueffed:
Adding a new field to an agent causes the endowment generator to pull an extra random value from the Random object. This causes the turn order to be different between versions.

nkremerh:
Fixed by 9351601.

nkremerh:
Previous solution was not sufficient. Trying again.

PR #8, opened 2023-07-12
Fixes determinism breaking when new agent fields are added

WillemHueffed:
Adds config hashes as fields to Sugarscape object
Adds child endowment hashes as fields to Agent object 
Modifies sugarscape.py to use the md5 hashing algorithm to select a number from random's stream 
Modifies agent.py to use the md5 hashing algorithm to to select a number from random's stream

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -1,3 +1,4 @@
+import hashlib
 import math
 import random
 import sys
@@ -60,6 +61,7 @@ class Agent:
         self.spicePrice = 0
         self.sugarPrice = 0
         self.nice = 0
+        self.childEndowmentHashes = None
 
     def addChildToCell(self, mate, cell, childConfiguration):
         sugarscape = self.cell.environment.sugarscape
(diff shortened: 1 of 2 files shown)


nkremerh (changes requested):
The functionality works as expected. Adding new fields to the config file (and using them in sugarscape.py) does not break determinism for previously added fields. Good work. Some style changes:

agent.py -
1. The hashlib import in agent.py should be moved to the top of the file.
2. Alphabetize the parentTraits (and rename that dictionary to parentEndowments, let's keep the terminology consistent).
3. Remove the setstate and getstate calls on lines 831, 833, 838, 880, and 884. These are redundant.

sugarscape.py -
1. Rename the hash dictionaries to agentConfigHashes and diseaseConfigHashes.
2. There's a newline on 283 that should be removed.
3. Rename penalties (line 358) to randomEndowments (or something similar) since not all disease features are penalties.
4. Alphabetize diseaseEndowment (lines 379-385).
5. Remove trailing comma on line 436.

config.json - 
1. Unstage this file change since there is no change.

Final step: get this peer reviewed for full functionality testing.

mmilkowski36:
Functionality looks fine.

nkremerh:
Fixes #7

PR #10, opened 2023-07-18
Removing dead code in findBestCell()

mmilkowski36:


nkremerh:
Something gone awry here with your rebase. Take a look at how many lines are being changed in the pull request. Nothing that can't be fixed with some Git magic.

PR #12, opened 2023-07-28
Adds scripts for data scraping json log files

WillemHueffed:
Modifies file hierachy, giving scripts and config files their own directories 
Modifies logparse.py to make import friendly


nkremerh:
Let's plan to have this fully ready to merge by next Tuesday's meeting. Follow up with me offline to give me a status update on where this PR is at and what is still left to be done to get it over the finish line.

WillemHueffed:
Save a final style review the PR should be good to go. It currently supports 9 graphs via "make plots" and "make clean".  Notes:
1. The shell script currently is hard coded to target '100-run-json/' - I know this is fine for us but should I have it read user input so people don't have to worry about modifying the script for their own data?
2. There is some code in parseLogsForGraphs.py that smells (lines 180-189 and 195-203). I'd like to throw these lines in a loop but a few of the graphs need special computation so I can't call the same function for all of them. 
2.1. On that note if we end up not needing the special graphs that need normalization by some other factor (eg agentWealth/envWealth) I can pretty easily modify the scraping script to scrape data for all model descriptors. Figuring out a way to have one master plotting script will be more challenging (largely because the legend needs moving to different quadrents depending on the shape of the graphs) however it may be worth while for simplicity of the file structure.
3. I flattened the file hierarchy and did my best to rename files to discussed conventions. However, I found my notes from out meeting less detailed than I would like. Please let me know if any of the naming conventions need changing.

PR #15, opened 2023-08-24
adds universal basic income

lucas-vorkpork:
@will can you peer review?

### DIFF
--- a/README
+++ b/README
@@ -219,6 +219,18 @@ agentTradeFactor: [float, float]
     Note: The more aggressive in trading an agent, the more resources they will attempt to trade.
     Default: [0, 0]
 
+agentUniversalSpice: [float, float]
+    Set the amount agents recieve for universal basic spice income.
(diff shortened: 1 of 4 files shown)


lucas-vorkpork:
Update: Don't pull, an error popped up in the process of squashing commits. Working on it

will (approved):
Lgtm

lucas-vorkpork:
Yeah, sorry about that, my mistake. I fixed up the error, too. @marishuman could you take a look at it?

nkremerh (changes requested):
It looks like the cause of the spiking metabolism is rampant disease (which increases agent metabolism). When comparing the same seed with and without UBI, the society without UBI overcomes disease and reaches a mean metabolism ~1.0 whereas the UBI population doesn't shake off disease and sees a skyrocketing metabolism (and the population dies out). Something to investigate.

nkremerh (approved):

PR #26, opened 2024-02-12
Consolidates redundant ethical score calculation code

colinhanrahan:
The only thing that changes between the 5 if statements in findBestEthicalCell is the actual ethics function being called. Everything else is the same, so it can be consolidated into one code block instead.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -619,44 +619,27 @@ class Agent:
         if self.decisionModel == "none":
             return greedyBestCell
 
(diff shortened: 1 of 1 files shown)


nkremerh:
Similar to another of your PRs, this is a great idea which we put off for awhile as we finished up papers. This functionality has now been implemented with each decision model having its own class (each a child of the Agent class). The redundant code is now gone, and there is now one less place needed to be edited when adding in new decision models.

PR #27, opened 2024-02-12
Removes duplicated addition of universal income every timestep

colinhanrahan:
self.sugar += self.universalSugar and self.spice += self.universalSpice are already present in doUniversalIncome() which is called in doTimestep(). Based on the time interval logic in doUniversalIncome(), it doesn’t seem like universalSugar and universalSpice should be added every single timestep in addition to the call to doUniversalIncome().


nkremerh:
@mmilkowski36 Please review.

PR #28, opened 2024-02-14
Replaces insertion sort and reverse with reverse order insertion sort

colinhanrahan:
Whenever sortCellsByWealth() is called, reverse() is called immediately after. Sorting the cells in reverse order in the first place is more efficient. This commit removes the sortCellsByWealth() function, replaces it with sortCellsByWealthDescending() (which sorts in descending order), and replaces two instances of calls to sortCellsByWealth() and then reverse() with sortCellsByWealthDescending().

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -611,7 +611,7 @@ class Agent:
         if len(cells) == 0:
             return None
         bestCell = None
-        cells = self.sortCellsByWealthDescending(cells)
+        cells = self.sortCellsByWealth(cells)
         if "all" in self.debug or "agent" in self.debug:
             self.printCellScores(cells)
         # If not an ethical agent, return top selfish choice
@@ -663,7 +663,7 @@ class Agent:
 
         # If additional ordering consideration, select new best cell
         if "Top" in self.decisionModel:
-            cells = self.sortCellsByWealthDescending(cells)
+            cells = self.sortCellsByWealth(cells)
             if "all" in self.debug or "agent" in self.debug:
                 self.printEthicalCellScores(cells)
             bestCell = cells[0]["cell"]
@@ -1219,7 +1219,7 @@ class Agent:
         sugarscape = self.cell.environment.sugarscape
         sugarscape.addDisease(disease, agent)
 
-    def sortCellsByWealthDescending(self, cells):
+    def sortCellsByWealth(self, cells):
         # Insertion sort of cells by wealth in descending order with range as a tiebreaker
         i = 0
         while i < len(cells):
--- a/agent.py
+++ b/agent.py
@@ -611,8 +611,7 @@ class Agent:
         if len(cells) == 0:
             return None
         bestCell = None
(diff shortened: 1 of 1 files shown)


goolabjamun:
Perhaps we can consider using the built-in Python sort routine or some other sort that operates in O(n log n) time instead of O(n^2) time.

colinhanrahan:
The lists passed in to this function are very small — In my tests, both insertion sort and quicksort take 0.000000 seconds. std::sort() in C++ actually prefers insertion sort for smaller lists because it's faster. @nkremerh I actually don't think sorting the cells is ever necessary in findBestEthicalCell (the only place the sort function is called). The Binary decision models iterate through the ethical values and find the first positive one anyway. (If there is a correlation between wealth and ethical value, removing the shuffle will change the behavior somewhat, but the decision is still naive and binary.) The Top models only need to find the best ethical value in the list, which can be achieved with a max rather than a sort.

nkremerh:
It is not strictly necessary, that's correct. However, the sorting places the options in greedy order (emulating the intended behavior of the original Sugarscape) then finds the first ethically acceptable option from that list. The intended effect is that the ethical decision model is placed _atop_ Sugarscape rather than _replacing_ things whenever possible.

colinhanrahan:
That makes sense. I was considering using a max for the "if Top" block instead of sorting, but it seems less readable than just calling the function again and its impact seems to be negligible. After testing, insertion sort seems appropriate for the small list size, and I renamed sortCellsByWealthDescending back to sortCellsByWealth in a commit on this branch. Should be OK to merge.

nkremerh (approved):

agent.py
```diff
@@ -1222,11 +1220,11 @@ def spreadDisease(self, agent, disease):
         sugarscape.addDisease(disease, agent)
 
     def sortCellsByWealth(self, cells):
-        # Insertion sort of cells by wealth with range as a tiebreaker
+        # Insertion sort of cells by wealth in descending order with range as a tiebreaker
         i = 0
         while i < len(cells):
             j = i
-            while j > 0 and (cells[j - 1]["wealth"] > cells[j]["wealth"] or (cells[j - 1]["wealth"] == cells[j]["wealth"] and cells[j - 1]["range"] <= cells[j]["range"])):
+            while j > 0 and (cells[j - 1]["wealth"] < cells[j]["wealth"] or (cells[j - 1]["wealth"] == cells[j]["wealth"] and cells[j - 1]["range"] >= cells[j]["range"])):
```
nkremerh, line 1227: One last adjustment: the inverse of less than or equal to is strictly greater than. So, a quick swap of >= to just > should have this ready to merge. Plus, it'll add stability to the sort which I unintentionally broke in the first place.

colinhanrahan: Good thinking, added in new commit.

nkremerh:
Good work!

PR #30, opened 2024-02-16
Moves instantiation of all cells from addSugarPeak() and addSpicePeak()…

colinhanrahan:
…to Environment constructor, removes now unused placeCell() definition.

The Environment constructor currently fills grid with Nones, and addSugarPeak() and addSpicePeak() both have checks to instantiate default Cells outside of the peak shape if self.environment.findCell(i, j) == None. addSugarPeak() and addSpicePeak() should only add sugar and spice peaks as it says in their names, not create the empty grid cells too. It doesn't make sense to replace grid Nones with Cells in both addSugarPeak() and addSpicePeak(). It's more intuitive to instantiate the Cell objects in the Environment constructor.

*I did have to import cell to environment, but I think that makes sense given that grid stores Cells.

### DIFF
--- a/environment.py
+++ b/environment.py
@@ -1,5 +1,6 @@
 import math
 import random
+import cell
 
 class Environment:
     # Assumption: grid is always indexed by [height][width]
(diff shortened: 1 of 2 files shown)


nkremerh:
Great catch! This inspired me to write an alternative implementation which has the same complexity class but is a bit cleaner.

PR #33, opened 2024-02-17
Fixes indentation and loop logic in findEthicalValueOfCell()?

colinhanrahan:
This change is purely based on the fact that `Egoist findEthicalValueOfCell()` is calling `neighbor` outside of the `for neighbor in self.neighborhood` loop, which doesn't seem right to me. I put the `canSee` sum in a separate loop because it seems like it's only supposed to be used in the main loop after it's fully calculated. `canSee` is used for calculating `extent` for each `neighbor`, but to be honest, I'm not sure what `extent` is doing inside of the loop. I thought it was indicated by the summing of each `neighborValueOfCell` into `cellValue`. `canSee` is one of the only differences between the `findEthicalValueOfCell()` in `Egoist` and the one in `Bentham`, which just uses `len(self.neighborhood)` to calculate `extent`.

### DIFF
--- a/ethics.py
+++ b/ethics.py
@@ -149,28 +149,28 @@ class Egoist(agent.Agent):
         for neighbor in self.neighborhood:
             if neighbor.canReachCell(cell) == True:
                 canSee += 1
-
-        for neighbor in self.neighborhood:
-            timestepDistance = 1
-            neighborMetabolism = neighbor.sugarMetabolism + neighbor.spiceMetabolism
-            cellDuration = cellSiteWealth / neighborMetabolism if neighborMetabolism > 0 else 0
-            certainty = 1 if neighbor.canReachCell(cell) == True else 0
-            proximity = 1 / timestepDistance
-            intensity = (1 / (1 + neighbor.findTimeToLive()) / (1 + cell.pollution))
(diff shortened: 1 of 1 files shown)


nkremerh:
The egoist model works differently because the agent really only cares about the consequences (so, utility) for itself. We figure out which neighbors can reach the cell under consideration (different from the size of the neighborhood) to figure out how many agents will be affected if the agent currently moving goes to this cell, but we don't care how those other agents will be affected. So, there is no indentation issue. Instead, it's a logical issue. All those references to `neighbor` should be references to `self` instead.

colinhanrahan:
That makes way more sense, thank you. It doesn't make sense for egoists to take others into consideration. I was comparing the code of the different `findEthicalValueOfCell()` functions to see what would need to happen for them to be merged into one that just uses `selfishnessFactor`. I'll fix the logic in this function and then maybe you can test to see if the behavior better matches a `Bentham` with a `selfishnessFactor` of 1.

colinhanrahan:
New commit should properly fix the function, but I might've missed a few neighbor-related calculations or specifics related to the utility equation.

nkremerh:
Deprecated by 82647de. Fixes to Egoist class here helped verify that using the Bentham class with an agent's `selfishnessFactor` set to 1 produces (nearly) the same behavior.

Issue #18, opened 2023-12-20
Agent movement is inefficient


nkremerh:
The bulk of runtime per agent is spent determining their movement each timestep. There seem to be multiple inefficiencies at play causing this search and move to consume a lot of cycles that should be diagnosed and patched. This will eliminate a bottleneck preventing simulation throughput during data collection.

nkremerh:
Partially addressed by #29. There are likely other inefficiencies slowing this down so drastically compared to the default simulation behavior.

nkremerh:
Performance gains made by #29, #36, and other recent code changes have made agent movement much more tolerable. Changing this to a low priority issue as there are likely still gains to be made which can be improved upon in a lazy, as-they-come-up nature.

nkremerh:
Further resolved by #83.

PR #36, opened 2024-02-19
Doubles efficiency of findCellsInRadialRange

colinhanrahan:
- Circles are symmetrical, so you only need to check the top left quadrant and then you can reflect those cells to the other three quadrants.
- I also included the current cell in `findCellsInCardinalRange` because it's included in `findCellsInRadialRange` and it seems like it should be in range of itself.

I figured radial movement and vision — especially at higher values — would be pretty time intensive for the simulation, so I wanted to make it as efficient as possible. This should help out with #18.

### DIFF
--- a/environment.py
+++ b/environment.py
@@ -84,7 +84,7 @@ class Environment:
                 self.grid[i][j].findNeighbors()
 
     def findCellsInCardinalRange(self, startX, startY, gridRange):
-        cellsInRange = []
+        cellsInRange = [{"cell": self.grid[startX][startY], "distance": 0}]
         for i in range(1, gridRange + 1):
             deltaNorth = (startY + i + self.height) % self.height
             deltaSouth = (startY - i + self.height) % self.height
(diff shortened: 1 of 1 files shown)


environment.py
```diff
@@ -84,29 +84,42 @@ def findCellNeighbors(self):
                 self.grid[i][j].findNeighbors()
 
     def findCellsInCardinalRange(self, startX, startY, gridRange):
-        cellsInRange = []
+        cellsInRange = [{"cell": self.grid[startX][startY], "distance": 0}]
```
nkremerh, line 87: Agents should not consider their current cell as a viable option. Sugarscape assumes agents _must_ move somewhere else with not moving be a fallback if there are truly no other options. Remove this line.

environment.py
```diff
@@ -84,29 +84,42 @@ def findCellNeighbors(self):
...
-            deltaWest = (startX - i + self.width) % self.width
-            cellsInRange.append({"cell": self.grid[startX][deltaNorth], "distance": i})
-            cellsInRange.append({"cell": self.grid[startX][deltaSouth], "distance": i})
-            cellsInRange.append({"cell": self.grid[deltaEast][startY], "distance": i})
-            cellsInRange.append({"cell": self.grid[deltaWest][startY], "distance": i})
+            deltaNorth = (startY + i + height) % height
+            deltaSouth = (startY - i + height) % height
+            deltaEast = (startX + i + width) % width
+            deltaWest = (startX - i + width) % width
+            cellsInRange.extend([
```
nkremerh, line 95: If this can cleanly fit on 1-2 lines, please compress.

environment.py
```diff
@@ -84,29 +84,42 @@ def findCellNeighbors(self):
...
-                    deltaX = (i + self.height) % self.height
-                    deltaY = (j + self.width) % self.width
-                    cellsInRange.append({"cell": self.grid[deltaX][deltaY], "distance": euclideanDistance})
+                    # Coordinates for the top left quadrant
+                    deltaX = (i + height) % height
+                    deltaY = (j + width) % width
+                    # Reflected coordinates for the other three quadrants
+                    reflectedX = (2 * startX - i + height) % height
+                    reflectedY = (2 * startY - j + width) % width
+                    cellsInRange.extend([
```
nkremerh, line 118: Same as above, compress if possible.

nkremerh (changes requested):
Pretty close to being ready for prime time. Seeing marginal improvement in runtime at 500 timesteps between this PR and the current repo head.

However, the results differ between the two even when using the same config and seed. This is due to the change on line 87. For each improvement, we need to ensure determinism is retained unless there's a good reason to break it.

environment.py
colinhanrahan: Fixed in new commit, but it's worth noting that the currently implemented behavior includes the current cell in `findCellsInRadialRange()`.

environment.py
colinhanrahan: Compressed in new commit, but I'm not sure if it's the style you're looking for. Let me know.

environment.py
colinhanrahan: Compressed in new commit, but I'm not sure if it's the style you're looking for. Let me know.

nkremerh: I think it reads better to have it expanded rather than in a list comprehension (so how you had it before) just not one on each line. See if it looks nice with 2 per line or all on 1 line.

environment.py
nkremerh: That should likewise be taken out. Good catch.

environment.py
colinhanrahan: 1 line looks too long to me. 2 is alright for the variables, but then you have to have 4 lines for the `extend()`.

colinhanrahan:
New commits reformat lines again and remove the current cell from findCellsInRadialRange. Let me know what you think.

PR #37, opened 2024-02-21
Improves overall simulation efficiency by 33%, fixes extent calculation

colinhanrahan:
To start off, half of the runtime on the current simulation (with ethical models) is taken up by calls to `self.findNeighborhood(cell)` in the `for neighbor in self.neighborhood` loop in `findEthicalValueOfCell()`. I noticed this does not need to be calculated every iteration of the loop — just once at the beginning. This would improve simulation efficiency by 50%. 

Then I noticed that the calculation it was being used in — `futureExtent` — was not updated to include radial movement and vision, and neither was `extent`. They should both be calculated by finding the proportion of the neighbor's neighborhood that is shared by the current agent's neighborhood (present or future). I changed `neighborhood` to be a set because it's much faster to find shared values in sets. Calculating the shared neighbors for each neighbor is still time intensive, but still much more efficient than the current code.

Overall, this will massively improve efficiency for any decision model that calls `findEthicalValueOfCell()`. In my test, a simulation with 200 timesteps using `benthamHalfLookaheadBinary` that originally took 300 seconds now takes 200 seconds, and all of the time reduction is in `findEthicalValueOfCell()`.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -47,7 +47,7 @@ class Agent:
         self.alive = True
         self.age = 0
         self.cellsInRange = []
-        self.neighborhood = []
+        self.neighborhood = set()
         self.lastMoved = -1
         self.vonNeumannNeighbors = {"north": None, "south": None, "east": None, "west": None}
         self.mooreNeighbors = {"north": None, "northeast": None, "northwest": None, "south": None, "southeast": None, "southwest": None, "east": None, "west": None}
@@ -84,6 +84,7 @@ class Agent:
         self.visionModifier = 0
         self.aggressionFactorModifier = 0
         self.fertilityFactorModifier = 0
+        self.findNeighborhood()
 
     def addChildToCell(self, mate, cell, childConfiguration):
         sugarscape = self.cell.environment.sugarscape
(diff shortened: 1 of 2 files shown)


colinhanrahan:
This should help a lot with #18. Also, if #36 is accepted, this commit should be changed to remove `neighborhood.add(self)` in `findNeighborhood()` as the cell/agent will already be added in `findCellsInCardinalRange()`.

colinhanrahan:
There is an issue with this: agents will only update their own neighborhood when moving, so any other agents' neighborhoods that the current agent moves in or out of will be incorrect for the rest of the timestep until their own `moveToBestCell()` is called. If this is not an acceptable error, then every agent whose neighborhood is affected by the move will need to have their neighborhood updated, which is likely pretty inefficient. The best way to do this would be to subtract the current agent from all of its neighbors' neighborhood sets before a move and add it back to all its new neighbors' neighborhood sets after the move. However, the current behavior doesn't seem to have any catastrophic effect on the behavior of the simulation

colinhanrahan:
Update:
- Before the move, you would only need to update the neighbor's neighborhood by subtracting the current agent if `neighbor.vision >= self.vision` (if the current agent is in the neighbor's neighborhood).
- After the move, you would need to update all possible neighborhoods that the agent could have moved into, which includes neighborhoods of agents outside of the current agent's neighborhood if `self.vision <= neighbor.vision`. This could be achieved by finding the neighborhood with the max possible vision and updating the neighborhood of every neighbor agent if `neighbor.vision >= self.vision`, but it would be very inefficient.

Issue #17, opened 2023-12-20
Long runs have a large memory footprint


nkremerh:
Running the simulation in excess of 1,000 timesteps (given a successful seed with a thriving population) leads to a large memory footprint (> 8GB). There is likely one or more significant memory inefficiencies at play that should be diagnosed and patched. This will eliminate a bottleneck preventing more simulations running in parallel during data collection.

colinhanrahan:
On my test with `benthamHalfLookaheadBinary`, at 1000 timesteps, there were 2 million dictionaries taking up half a gigabyte of memory. The memory graph is consistently going up. Looks like they aren't being garbage collected — I'll look into this further.

colinhanrahan:
Preliminary results: it seems to be that the social networks are keeping references to dead agents, which still have their own social networks which keep references, etc. so they build up over time. I removed everything from dead agents' social networks except children and debtors (which are still needed) and the number of dictionaries at 1000 timesteps went from 2 million to 350,000. I still need to check to see if the simulation behaves the same. I'm going to do a small overhaul of interaction between dead agents and social networks and see what happens.

colinhanrahan:
Update: The memory leak is mostly, solved, but the memory usage is still slowly increasing over time. Most, if not all, of the remaining dictionaries are loans whose references are somehow avoiding removal.

nkremerh:
Solved by #39.

PR #39, opened 2024-02-27
Fixes memory leak by removing old references (first pass)

colinhanrahan:
memory-profile
memory-profile-3

#17 
Before and after of a memory graph for `benthamHalfLookaheadBinary` over 1K timesteps; it appears to almost flatten out at under 150 MiBs now. I can't test thoroughly over 10K timesteps or multiple models and seeds on my computer, so maybe @nkremerh can see what it looks like.

Some notes:
- I'm not sure what the memory spike at the end is — maybe the garbage collector kicking in, or something related to the memory profiling?
- This fix does appear to slow the simulation down a bit, which is unfortunate. The graph appears to show 120 vs. 220 seconds on my test, which is surprising given how little code is added.
- From my investigation, there are still some objects that shouldn't be in memory anymore, but I think they're now being capped by the population so they're not causing memory leaks.
- I'm certain this maintains determinism; I checked multiple timesteps later in the simulation and everything was the exact same. However, every so often, the statistics at the top will be slightly different, but then they regain parity in the next timestep. I'm not sure what's causing this behavior.
- There is probably a way to make this fix cleaner and better at cleaning up memory. This is a first pass.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -211,6 +211,13 @@ class Agent:
         self.resetCell()
         self.doInheritance()
 
+        # Remove unnecessary references
+        self.socialNetwork = {"debtors": self.socialNetwork["debtors"],
+                              "children": self.socialNetwork["children"]}
+        self.neighborhood = []
+        self.vonNeumannNeighbors = {}
+        self.mooreNeighbors = {}
+
     def doDisease(self):
         diseases = self.diseases
         for diseaseRecord in diseases:
--- a/agent.py
+++ b/agent.py
@@ -211,9 +211,8 @@ class Agent:
         self.resetCell()
         self.doInheritance()
 
-        # Remove unnecessary references
-        self.socialNetwork = {"debtors": self.socialNetwork["debtors"],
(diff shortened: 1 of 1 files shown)


nkremerh:
This is a good catch! It's kind of cool that we were accidentally keeping a historical lineage of every living agent, but that's absolutely brutal for performance.

My initial testing is not showing the extended runtime issue you are encountering. I will update with more details when I run more extensive tests.

colinhanrahan:
Yeah, you could probably do some cool family tree visualizations or quantify how much inbreeding happens in a small population. But the circular references were causing the agents to not be garbage collected. I'm glad you're not encountering the same runtime issues I was — I only ran one test.

nkremerh:
In further head-to-head testing, this patch no only drastically reduces memory usage but also seems to marginally increase runtime (presumably there was some hidden overhead to managing the shared, gargantuan social network). Pushing a small change then merging. Good work!

PR #41, opened 2024-02-28
Removes unused code for agent finding best cell

colinhanrahan:
You can go commit by commit if you want a detailed breakdown. The main changes I made were removing some unused variables and adding a return case to the beginning of `findBestCell()` so that you can initialize the best values outside of the loop rather than in an if statement in every iteration of the loop.


nkremerh:
Looks promising. I'll likely push a couple formatting nitpicks, but I'd like you to do the correctness testing yourself here. Compare the results of running a handful of seeds (say: 12345, 54321, 98765, 56789, and 10101) with the main three decision models (benthamHalfLookaheadBinary, egoisticHalfLookaheadBinary, and rawSugarscape) out to 1k timesteps. Leave everything else in the config file as is.

If the results of the current head of the master branch and your change are the exact same for all cases, it's a perfect match! If you're wanting to run this on a machine that isn't your laptop, the CS1 server can run Sugarscape so long as `headlessMode` is set to `true`.

colinhanrahan:
I changed initialization of best values to what it was before (the if statement in the for loop is still not necessary) and behavior is now identical to current behavior over multiple models and seeds. Should be OK to merge.

One thing I've been noticing (on the current repo, not the PR) is that the statistics at the top are not always deterministic — they are very close but slightly off sometimes, especially closer to the beginning of simulations. For example, there will be 50 agents on screen and it will say there's 52. Any idea why this is, or particularly why it doesn't seem deterministic?

PR #43, opened 2024-03-02
Adds correct timesteps to make data simulations

colinhanrahan:
I noticed that my `make data` log files were blowing past the `plotTimesteps` I had set for them. It looks like the intended timesteps from `dataCollectionOptions` are not included in the simulation setups, so they use the default 1000.

As a side note, it might be faster to only log the variables you're intending to plot rather than log all of them and only use the `plots` setting to generate specific plots after the fact. It looks like all variables are being logged right now regardless of whether you're planning to plot them or not. I could fix this if it's a topic of interest.

Also, I had to manually change the `make data` command to python3 instead of python. Should the `make data` command be using the same python alias for running `run.py` that you can set for actually running the simulations in `dataCollectionOptions`?

### DIFF
--- a/data/run.py
+++ b/data/run.py
@@ -13,6 +13,7 @@ def createConfigurations(config, path):
         if path[-1] != '/':
             path = path + '/'
         dataOpts = config["dataCollectionOptions"]
+        timesteps = dataOpts["plotTimesteps"]
         seeds = generateSeeds(dataOpts)
         confFiles = []
         for seed in seeds:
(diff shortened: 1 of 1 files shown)


nkremerh:
Not an issue: that option is the number of timesteps to include in the generated plots, not how many timesteps the simulation should run. There may be good reason to collect lots of data (for example, 50k timesteps) but only plot the first 10k if the interesting behavior happens then. Or, if there is some behavior that really requires us to zoom in on the first 1k timesteps, for instance. I will add a note in the README clarifying this.

Adding specific data to the log based on the provided plots seems like a good idea on first blush, but it is common for us to not exactly know _what_ we want to plot at the moment of data collection. We can't know the interesting behavior until we see it happen. Hypothesizing too much beforehand can cause an issue down the road. It is better to err on the side of caution and collect a treasure trove which can be plumbed later rather than have too little and have to re-run a potentially lengthy data collection pipeline.

You are correct on the Python alias. That can be automatically set for you in the Makefile and in the config by running `make setup`. This is a recent commit, so if there is an issue with that please raise it as an issue.

Issue #19, opened 2023-12-20
bentham*LookaheadTop and egoist*LookaheadTop broken


nkremerh:
The behavior of the `bentham*LookaheadTop` and `egoist*LookaheadTop` decision models appears broken on first glance. Any model ending in `Top` should be selecting the locally optimum choice, yet this seems to always result in quick societal death during data collection (while the more naive `Binary` models do not). There is likely a logic bug that needs to be diagnosed and patched. Another avenue worth investigating is if the configuration options used during data collection prevent the `Top` models from making good choices.

nkremerh:
Maybe fixed due to many code changes to agent decision model behavior. Will need to verify.

colinhanrahan:
I had a hard time reproducing this as well.

nkremerh:
Nevermind. When running the simulation out to 750 timesteps over 20 seeds, I get the following report:

```
        Decision Model         Died  Worse Better
    benthamNoLookaheadTop     :  20     0     0  
    egoisticNoLookaheadTop    :  20     0     0  
   benthamNoLookaheadBinary   :   0     0    20  
  egoisticNoLookaheadBinary   :  12     0     8  
        rawSugarscape         :  12     0     8
```

nkremerh:
Same configuration while using the half lookahead decision models:

```
        Decision Model         Died  Worse Better
   benthamHalfLookaheadTop    :   1     0    19  
   egoisticHalfLookaheadTop   :   6     0    14  
  benthamHalfLookaheadBinary  :   0     0    20  
 egoisticHalfLookaheadBinary  :  10     1     9  
        rawSugarscape         :  15     0     5
```

colinhanrahan:
The only difference I can find for 'Top' models is the 'Top' check and subsequent sort in 'findBestEthicalCell()', which looks fine. My best guess is that some part of the ethical calculation is incorrect for certain cells. For example, maybe duration is incorrect for neighbors with 0 metabolism, or intensity is incorrect for really old agents. As an effect of this, a few cells are incorrectly calculated to have really high ethical scores, so they are chosen as the best cells when we are getting the best value for 'Top'. For binary models, it'll usually choose a correctly calculated cell by chance.

I'm not familiar enough with the ethical formula as implemented in the repo to check if these are all correct, but maybe other people working on the issue will find this useful.

colinhanrahan:
[...]

PR #44, opened 2024-03-03
Fully implements option to use Moore neighborhoods

colinhanrahan:
... as discussed in page 39, footnote 28 of Growing Artificial Societies.

Currently: agents have `self.mooreNeighbors` and `self.vonNeumannNeighbors`.
- `mooreNeighbors` is unused
- `vonNeumannNeighbors` is only used for `updateSocialNetwork()`
- direction info from dictionary like "north", "southwest" is never used

Currently: cells have `self.neighbors` (von Neumann). This is used for:
- `doReproduction()`
- `doDisease()`
- `doTagging()`
- `doTrading()`
- `findEmptyNeighborCells()`

I've updated the code to allow all of these use cases to use either Moore or von Neumann neighbors, all based on a new `adjacencyMode` config setting. (This could be updated to allow different adjacency modes for each use case). `adjacentNeighbors` now stores the Moore or von Neumann-adjacent agents for each agent, and `adjacentCells` stores the full Moore or von Neumann neighborhood of cells for each cell. I tried to make the names more distinguishable from `neighborhood` because they sound pretty similar.

I worked on this all at once so there's mostly just one commit. Here are a few big things I changed:
- von Neumann neighborhoods are now calculated using `findCellsInCardinalRange()` with a range of 1.
- Moore neighborhoods are now calculated using `findCellsInRadialRange()` with a range of 1.
- The directions stored in agent `self.*Neighbors` are now unused, so I've removed them altogether and replaced the data structure with a list of agents.
- Following this, the 4 `find[Direction]Neighbor()` functions are deprecated and have been removed.
- Included backwards compatibility for old config files; default adjacency mode is von Neumann.

I'll do my best to add comments on the commit where necessary. Simulations do not have any catastrophic changes and in fact appear normal, but I need to test more to make sure determinism is maintained for default von Neumann-based simulations.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -356,51 +356,55 @@ class Agent:
         # Agent marked for removal or not interested in reproduction should not reproduce
         if self.alive == False or self.isFertile() == False:
             return
-        random.shuffle(self.adjacentNeighbors)
+        random.shuffle(self.cell.adjacentCells)
         emptyCells = self.findEmptyNeighborCells()
-        for neighbor in self.adjacentNeighbors:
-            neighborCompatibility = self.isNeighborReproductionCompatible(neighbor)
-            emptyCellsWithNeighbor = emptyCells + neighbor.findEmptyNeighborCells()
-            random.shuffle(emptyCellsWithNeighbor)
-            if self.isFertile() == True and neighborCompatibility == True and len(emptyCellsWithNeighbor) != 0:
-                emptyCell = emptyCellsWithNeighbor.pop()
-                while emptyCell.agent != None and len(emptyCellsWithNeighbor) != 0:
+        for neighborCell in self.cell.adjacentCells:
+            neighbor = neighborCell.agent
+            if neighbor != None:
+                neighborCompatibility = self.isNeighborReproductionCompatible(neighbor)
+                emptyCellsWithNeighbor = emptyCells + neighbor.findEmptyNeighborCells()
+                random.shuffle(emptyCellsWithNeighbor)
+                if self.isFertile() == True and neighborCompatibility == True and len(emptyCellsWithNeighbor) != 0:
                     emptyCell = emptyCellsWithNeighbor.pop()
-                # If no adjacent empty cell is found, skip reproduction with this neighbor
-                if emptyCell.agent != None:
-                    continue
-                childEndowment = self.findChildEndowment(neighbor)
-                child = self.addChildToCell(neighbor, emptyCell, childEndowment)
-                self.socialNetwork["children"].append(child)
-                childID = child.ID
-                neighborID = neighbor.ID
-                self.addAgentToSocialNetwork(child)
-                neighbor.addAgentToSocialNetwork(child)
-                neighbor.updateTimesVisitedWithAgent(self, self.lastMoved)
-                neighbor.updateTimesReproducedWithAgent(self, self.lastMoved)
-                self.updateTimesReproducedWithAgent(neighbor, self.lastMoved)
-                self.lastReproduced += 1
+                    while emptyCell.agent != None and len(emptyCellsWithNeighbor) != 0:
+                        emptyCell = emptyCellsWithNeighbor.pop()
+                    # If no adjacent empty cell is found, skip reproduction with this neighbor
+                    if emptyCell.agent != None:
+                        continue
+                    childEndowment = self.findChildEndowment(neighbor)
+                    child = self.addChildToCell(neighbor, emptyCell, childEndowment)
+                    self.socialNetwork["children"].append(child)
+                    childID = child.ID
+                    neighborID = neighbor.ID
+                    self.addAgentToSocialNetwork(child)
+                    neighbor.addAgentToSocialNetwork(child)
+                    neighbor.updateTimesVisitedWithAgent(self, self.lastMoved)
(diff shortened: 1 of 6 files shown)


agent.py
```diff
@@ -49,8 +49,7 @@ def __init__(self, agentID, birthday, cell, configuration):
         self.cellsInRange = []
         self.neighborhood = []
         self.lastMoved = -1
-        self.vonNeumannNeighbors = {"north": None, "south": None, "east": None, "west": None}
-        self.mooreNeighbors = {"north": None, "northeast": None, "northwest": None, "south": None, "southeast": None, "southwest": None, "east": None, "west": None}
+        self.adjacentNeighbors = []
```
colinhanrahan, line 52: `adjacentNeighbors` will either store Moore or von Neumann neighbors depending on `adjacencyMode` config option.

agent.py
```diff
@@ -361,57 +357,51 @@ def doReproduction(self):
         # Agent marked for removal or not interested in reproduction should not reproduce
         if self.alive == False or self.isFertile() == False:
             return
-        neighborCells = self.cell.neighbors
-        random.shuffle(neighborCells)
+        random.shuffle(self.adjacentNeighbors)
         emptyCells = self.findEmptyNeighborCells()
```
colinhanrahan, line 361: Everything changed below is just indentation differences from a removed if statement.

agent.py
```diff
@@ -422,7 +412,8 @@ def doTimestep(self, timestep):
             self.lastMoved = self.timestep
             self.doUniversalIncome()
             self.moveToBestCell()
```
colinhanrahan, line 414: The old `updateNeighbors()` updated both `vonNeumannNeighbors` and `mooreNeighbors`, and then called `updateSocialNetwork()`. I've changed it to update `adjacentNeighbors` based on the structure of `adjacentCells` (which will either be Moore or von Neumann neighborhoods set at the beginning of the simulation). `updateSocialNetwork()` has been moved outside this function as its own part of `doTimestep()`.

cell.py
```diff
@@ -43,55 +43,12 @@ def doSugarProductionPollution(self, sugarProduced):
         productionPollutionFactor = self.environment.sugarProductionPollutionFactor
         self.pollution += productionPollutionFactor * sugarProduced
```
colinhanrahan, line 45: the old `findNeighborAgents()` has been incorporated into `updateAdjacentNeighbors()`, and the old `findNeighbors()` functionality has been moved to `findVonNeumannNeighbors()` in `environment.py`.

environment.py
```diff
@@ -114,6 +129,14 @@ def findCellsInRadialRange(self, startX, startY, gridRange):
                     cellsInRange.append({"cell": self.grid[reflectedX][reflectedY], "distance": euclideanDistance})
         return cellsInRange
```
colinhanrahan, line 131: The two `findCellsIn*Range()` functions return lists of dictionaries with `cell`s and `distance`s, so the `cell`s are extracted into a list for the adjacency neighborhood.

sugarscape.py
```diff
@@ -34,11 +34,13 @@ def __init__(self, configuration):
                                     "universalSugarIncomeInterval": configuration["environmentUniversalSugarIncomeInterval"],
                                     "sugarscapeSeed": configuration["seed"]}
         self.seed = configuration["seed"]
```
colinhanrahan, line 36: Moved initialization of `self.debug` above initialization of `environment` so that the `Environment` can have its own `self.debug`.

sugarscape.py
```diff
@@ -34,11 +34,13 @@ def __init__(self, configuration):
...
+        self.debug = configuration["debugMode"]
         self.environment = environment.Environment(configuration["environmentHeight"], configuration["environmentWidth"], self, environmentConfiguration)
         self.environmentHeight = configuration["environmentHeight"]
         self.environmentWidth = configuration["environmentWidth"]
-        self.configureEnvironment(configuration["environmentMaxSugar"], configuration["environmentMaxSpice"])
-        self.debug = configuration["debugMode"]
+        # Maintain backwards compatibility for configs without adjacencyMode
+        adjacencyMode = configuration.get("adjacencyMode", None)
```
colinhanrahan, line 42: `get()` sets `adjacencyMode` to whatever it is in the config settings, or to a default value of `None` if there's no `adjacencyMode` in the config settings.

colinhanrahan:
This looks pretty good to me, but the original reason I wanted this feature was to see if I could fix the population issues with `*Top` models (see my comments on #19). `doReproduction()` and all the other functions I mentioned above still use `cell.neighbors` (which is von Neumann) instead of the agent's adjacency neighborhood, so you should make those functions use the agent's neighborhood mode if we're following the same line of thinking on the intended behavior for this feature.

nkremerh:
You're right, I missed that! It's been patched in. Good catch.

colinhanrahan:
Looks good to me! The only other thing we might consider adding is backwards compatibility for the neighborhood mode setting when verifying the config. It'll be interesting to see what effects this feature has on our simulations.

nkremerh:
If the config file doesn't specify the option, it defaults to `vonNeumann` in the config inside `sugarscape.py`, so there shouldn't be a need to handle anything (yet) for compatibility.

PR #45, opened 2024-03-11
Adds negative utilitarianism

hadiyachishti:


nkremerh:
This is a good first step. I'd like to see the core pieces of this approach integrated within the `findEthicalValueOfCell` method in `ethics.py` instead of on its own.

negative utilitarianism
```diff
@@ -0,0 +1,38 @@
...
+            positiveCellValue += neighborValueOfCell
+        
+        if selfishnessFactor != -1:
+            if neighbor == agent:
+                positiveCellValue *= selfishnessFactor
+            else:
+                if neighborValueOfCell < 0:
+                    negativeCellValue *= (1 - selfishnessFactor)
+
+    if positiveCellValue >= abs(negativeCellValue):
```
nkremerh, line 35: This function should return _both_ the positive and negative values of the cell. This is to resolve ties where two potential cells have the same negative value (as in, they case the same amount of harm) in which case the cell with the best positive consequences should be preferred.

This comparison of different cells needs to take place outside this function (and within `agent.py`) which is why returning a tuple or list of both is necessary.

PR #48, opened 2024-03-23
Removes agent sex info from GUI

colinhanrahan:


nkremerh:
We're running into a space limit here: long decision model names can take the text over the boundary of the GUI.

After thinking about this a bit, I realize that there is already color coding for showing decision models. Instead of adding this in, I'd actually like you to flip this PR to take agent sex out of the GUI info string as well (since the user can also color code the interface for that). The agent info string should only have info that can't be easily conveyed by changing the agent's coloration in the GUI.

Perhaps that will free up some space for additional info in the future (but let's not make any predictions about which properties may be relevant right this second).

colinhanrahan:
Removed model and sex from GUI in new commit. Since Sugarscape is generally a square environment and doesn't fill  screens anyway, we can always add more info on the left or right side if space is an issue.

PR #49, opened 2024-03-23
Removes unintended trait correlation and unused variable

colinhanrahan:
Current repo behavior:
- `pairedEndowmentIndex` is never used and `if... endowment in pairedEndowments` will never run because `parentEndowments` and `pairedEndowments` are mutually disjoint.
- The `pairedEndowments` loop uses `index`, which is the same as it was on the last iteration of the `parentEndowments` loop, which correlates traits that should not be correlated.
- Three separate hashes are created for `pairedEndowments`, but only one is needed because they all use the same value.

Updated behavior:
- Only one hash is created for `pairedEndowments`
- All `pairedEndowments` use the same value, which is now independent of the last iteration of the `parentEndowments` loop

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -721,20 +721,20 @@ class Agent:
                 hashed = hashlib.md5(config.encode())
                 hashNum = int(hashed.hexdigest(), 16)
                 self.childEndowmentHashes[config] = hashNum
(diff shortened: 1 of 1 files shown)


nkremerh:
Closing for now.

PR #50, opened 2024-04-26
adds negative utilitarianism

hadiyachishti:


agent.py
```diff
@@ -607,6 +607,8 @@ def findBestCell(self):
             print("Agent {0} moving to ({1},{2})".format(self.ID, bestCell.x, bestCell.y))
         return bestCell
 
+
+
```
colinhanrahan, line 623: I think this extra whitespace can be removed.

colinhanrahan (commented):
Sorry for the late review—I left these comments a while ago but apparently when they're "Pending" they're only visible to me.

colinhanrahan:
Hey @hadiyachishti, I noticed that you're adding a lot of files to the commit and then removing them. If you only want to add tracked files (the code you're changing), you can use `git add -u` or run `git status` to see what you changed and then `git add <file>` one by one. I use `git status` all the time to make sure I'm staging the correct files.

Issue #51, opened 2024-05-04
UBI placement during an agent's timestep leads to bugs


nkremerh:
Issue is hotfixed in cd26c63.

Depending on the placement of the `doUniversalBasicIncome` method in the agent's `doTimestep` method, some weird behavior either occurs or never shows. When UBI is handed out before `moveToBestCell`, the bugged behaviors below occur. When UBI is handed out after `moveToBestCell`, there are no bugged agents.

Agents are sometimes choosing their own cell as the best place to move (which is only allowed as a last resort but was happening when other choices existed). Agents are sometimes getting evicted from their cell while still being alive (allowing the population to grow past the number of cells).

Simply moving the UBI to _after_ `moveToBestCell` but _before_ `doMetabolism` (and the rest of the timestep) was sufficient for fixing the behavior without breaking UBI.

The root cause for this bug is likely to be some edge case(s) in the `moveToBestCell` method or other methods it calls. It is likely getting stress tested by UBI in ways not present with the other simulation options. Fixing the root cause (even if it _is_ in UBI) will close up any edge cases which may affect future features.

colinhanrahan:
For more context, what does "eviction" mean? Are agents disappearing from the map, or are there sometimes multiple agents in a cell, or something else?

nkremerh:
An agent has a reference to a cell in the environment. A cell likewise has a reference to the agent currently at that cell.

Eviction here means an agent moves to a cell in a timestep, but at some point (either before the conclusion of their turn or by the time it is their turn again) the reference to that agent is lost at the cell. The agent is still alive and may move to a new cell since it still has a reference to their old cell (i.e. `agent.cell != None` but `cell.agent == None` or `cell.agent ==` some other agent).

This means the total population can grow beyond the hard limit of the number of cells in the environment. Additionally, evicted agents are safe from any combat and cannot participate in trade, lending, or reproduction until their next move.

colinhanrahan:
Update: I've found very few evictions, but every one that I have found is related to a diseased agent. I couldn't find a single eviction in 25 seeds with UBI on and disease off. Maybe we are only seeing more evictions with UBI because UBI influences the spread of disease?

nkremerh:
You're likely right that this has something to do with disease and UBI making endemic disease an almost sure thing when the UBI is high enough.

[...]

PR #52, opened 2024-05-08
Fixes eviction bug

colinhanrahan:
For #51, agents doing the evictions ("evictors") always have negative aggression due to disease modifiers. Unlike `findMovement()`, `findVision()`, `findSugarMetabolism()`, and `findSpiceMetabolism()` (other values affected by disease modifiers), `findAggression()` does not account for negative values by using a floor value of 0. Negative aggression values cause undefined behavior, resulting in evictions. Updating this function to match the others appears to entirely remove eviction bugs.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -555,7 +555,7 @@ class Agent:
             return agent.wealth
 
     def findAggression(self):
-        return self.aggressionFactor + self.aggressionFactorModifier
+        return max(0, self.aggressionFactor + self.aggressionFactorModifier)
 
     def findBestCell(self):
         self.findNeighborhood()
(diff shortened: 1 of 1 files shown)


colinhanrahan:
For more detail: I believe `isNeighborValidPrey()` is calling `findAggression()`, which returns a negative value, and then deciding that a neighbor is valid prey because it checks `if aggression == 0` rather than `if aggression <= 0`. No neighbor should be valid prey with a negative aggression value. This if statement could be changed for even more security.

nkremerh:
Good find! I had a feeling it was something small.

Add in that patch to that check as well.

colinhanrahan:
Done!

nkremerh:
That should be fine. Do a check to make sure that doesn't break anything, then I'll merge it.

colinhanrahan:
Looks good to merge.

Issue #54, opened 2024-05-10
Book examples need to be tested and updated to current config file specification


nkremerh:
The configs in the `examples` directory demonstrate the behaviors present in the Growing Artificial Societies book. Each one needs to be tested to make sure it still works. This could also uncover bugs in the software that are not apparent when using the full set of options the default `config.json` enables.

There is already one bug apparent from the `agent_replacement.json` config file which causes a break in `agent.py`.

PR #55, opened 2024-05-11
Checks all of our examples against the book's diagrams and updates some examples

colinhanrahan:
I checked the first 6 examples for #54 and encountered some problems unrelated to the issue. I'd like some feedback before continuing. 

`agent_replacement.json`: Fixed break when all potential cells are skipped over due to conditionals. This is the only true error/crash I've encountered so far. All other commits in this PR are attempts to make the examples more closely follow the book.

`blank.json`: Looks good.

`combat_limited.json`: Added agent replacement as directed on page 86.
- The two tribes are supposed to be in separate quadrants at the beginning, not mixed together. Implementing this requires either a new config setting or a hardcoded check for the JSON name. Let me know which one you prefer.

`combat_unlimited.json`: Removed reproduction to follow the ruleset on pages 84-85.
- Same tribe separation problem as above.
- I noticed here that environment wraparound is not present in many of the book simulations, and actually takes away from the “battlefronts” described in this example. It would be a nontrivial task, but we could implement a config rule determining wraparound behavior. (This could be different per axis if needed: “x”, “y”, “both”, “none”.)
- The pictures on page 84 make no sense to me. The entire red population dies out much earlier than the blue population, but there's no precedent for this in the ruleset.

`constant_growback.json`: Looks good.

`cultural_tagging.json`: Changed number of tribes from 3 to 2 to follow main example instead of footnote on page 73. 
- The pictures on page 75 don’t make sense to me. There’s not supposed to be any replacement or reproduction in this example, but the agent count remains relatively stable across the 6 pictures. Agents should be starving and dying of old age. Either cultural transmission is happening extremely fast or agents are somehow not starving or getting old. Attempting to fix this with infinite agent lifespan does not resemble the pictures at all. The book discusses running this example for 3000 timesteps, so I’m not sure what is supposed to be happening here.


nkremerh:
I'm glad the bug was something small (it usually is).

On the combat examples, adding a new mechanism to have a given quadrant only be populated by one tribe (or a restricted list of tribes) is something worth thinking about. I've taken a couple passes at this but lost motivation before completing anything. If this is something you'd like to think about, let's talk about a potential implementation sometime. Let's not hold up this PR on this feature.

Undo the changes on the cultural transmission since going based purely off what the book gives us would obviously lead to a dead society by the end of the simulation. Reproduction is necessary (replacement would throw off the cultural osmosis happening) to get behavior approaching what the book shows. I also like the three tribes being included because it demonstrates our software can not only do the bare minimum of what's in the main text but also the enhancements the authors put in the footnotes.

To your confusion about a couple examples (such as the seeming lack of wraparound on the torus), I agree that is puzzling. One thing to consider is that we're only seeing snapshots. It could be that the few agents who have wrapped around were killed _off-screen_ so to speak. Or, they would have to go through a small sugar desert when they could instead move toward the center where there is a relatively fertile valley _and_ with combatants to plunder. Either is possible, but we'll never know for certain. What we have is a good faith reproduction of the behavior with more transparency in how it all happens.

Good work!

colinhanrahan:
Closing non-critical updates for now, see #57 for break fix.

nkremerh:
This should be re-opened when ready to roll back the cultural transmission changes. Everything else is good to go.

colinhanrahan:
Changes to `cultural_tagging.json` have been rolled back. To be clear, the other changes are still in place, even though they may make the examples less successful. For example, `combat_unlimited` is not supposed to include reproduction. I will look through the rest of the examples when I have time.

colinhanrahan:
Changes to `cultural_tagging.json` have been rolled back. To be clear, the other changes are still in place, even though they may make the examples less successful. For example, `combat_unlimited` is not supposed to include reproduction. I will look through the rest of the examples when I have time.

colinhanrahan:
I just finished looking through all of the examples. All of them are a good match for the book unless I indicate otherwise. Here are my results:

`immediate_growback.json`
- Added infinitely-lived agents in accordance with page 25: “If an agent does not starve it lives forever.”

`agent_replacement.json`
- Changed `agentStartingSugar` to `[5, 25]` following page 33.

`seasonal_migration.json`
- Seasonal migration is more accurately pages 44-46 (rather than 43-45), so I changed these numbers in the example and the README. This might help avoid any confusion between the diagonal migration waves on page 43 (which are not season-based) and seasonal migration.
- Age is not specified, so I chose infinite lifespans to better demonstrate the long season cycles.

`pollution_formation.json`
- Looks relatively close to the diagram, but the book specifies directions that we don't have: “At t = 50 pollution begins (rule P11 is turned on). Then, at t = 100, diffusion begins (rule D1 is switched on).” This could be a config option.

**`cultural_tagging.json`**
- See #64. Current behavior is not a very good match for the book.

**`combat_limited.json`**
- Doesn't work at all.
- Requires the tribal starting quadrants feature that we've talked about a bit, and likely a wraparound toggle in the config (which is on the Trello board). I've worked on both of these but will hold off on them until we make some more progress on other more important things.

**`disease_basic.json`**
Our disease config options do not fully encompass the book's directions:
- For Animation V-1, each initial agent has a random 4 of 10 diseases.
- For Animation V-2, each initial agent has a random 10 of 25 diseases.

This should be a separate PR. It should just be adding a config option called something like `diseasesPerStartingAgent`.
_However_, the behavior we can see here is actually much more compelling and interesting than the book. There are two population centers, one of which gets cured of diseases and another one that has many of them. When they collide, waves of disease spread through the population with weak immune systems. Check it out!

`constant_growback.json`, `reproduction_basic.json`, `reproduction_limited.json`, `reproduction_inheritance.json`, `combat_unlimited.json`, `spice_growback.json`, `trade_basic.json`, `trade_replacement.json`, `trade_pollution.json`, `foresight_basic.json`, and `lending_basic.json` all behave normally as far as I can tell.

We could add examples for pages 60, 62, and 65 (and others I may have missed). In particular, we should probably add an example demonstrating diagonal waves migration on page 43, which is one of the most interesting emergent behaviors in the book. See the Trello board for our progress on this.

The natural next step is to reproduce all of the plots and graphs in the book (not just the simulation diagrams). This is a longer project and accurately displaying all of the different graphs will require a large chunk of code.

colinhanrahan:
Merged some changes from master. One of our commits since May 10 changes`disease_basic.json` from the interesting behavior we found here to a much more boring simulation. (@nkremerh, my visualization branch maintains determinism with the repo—it's this branch that was deviating).

PR #59, opened 2024-05-26
Adds agent tracking and visual highlighting to GUI

colinhanrahan:
I've always wanted to be able to track agents across timesteps, so I added it. If you click on an agent, it'll track the agent until it dies, and if you click on a cell, it'll track the cell like current repo behavior. I added some visual highlighting — you can click on the same cell again to deselect it. I also simplified the strings using `.format()` and some redundant coordinate calculations for better readability. Let me know what your thoughts are!

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -19,9 +19,9 @@ class GUI:
         self.widgets = {}
         self.lastSelectedAgentColor = None
         self.lastSelectedEnvironmentColor = None
-        self.lastSelectedCell = None
-        self.selectedAgent = None
-        self.selectedAgentRectangle = None
+        self.highlightedCell = None
+        self.highlightedAgent = None
+        self.highlightRectangle = None
         self.activeColorOptions = {"agent": None, "environment": None}
         self.menuTrayColumns = 4
         self.menuTrayOffset = 110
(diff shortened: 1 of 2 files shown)


colinhanrahan:
Pushing some more changes for testing:
- Moves info bar to right side.
- Locks environment aspect ratio (cells will always stay square).
- Initializes runtime stats at timestep 0. Previously, they would be empty until timestep 1 unless you clicked on a cell, which would call `doTimestep()` and update them. This is the single change to `sugarscape.py`.
- Increases size of automatic window sizing. Quarter-window sizing looks a little too small, especially on my laptop.

I'm not sure how it looks outside of my Chromebook screen, so let me know when you get the chance.

colinhanrahan:
Updated following discussion this morning. Left click maintains previous functionality, and right click tracks agents. You can left or right click twice to deselect. Looks good to pull, but let me know if it doesn't work for you or if there's anything else I can add.

colinhanrahan:
It seems like `Button-3` is right click on my laptop but middle click on my Mac, so I guess `event.num` values differ across different input devices. I don't think there's a perfect solution for this, but agent tracking could be reconfigured to use a keyboard button instead.

nkremerh:
How about a double click?

colinhanrahan:
If you have a bind for a single click and a double click on the same key, both bindings will be called. I could hack my way around this by making my own double click logic instead of binding to `<Double-Button-1>`, or I could change it to a button and a modifier, like `<Control-Button-1>`.

colinhanrahan:
@nkremerh Added control click and removed commented code.

nkremerh (approved):

PR #61, opened 2024-06-01
Fixes index mismatch between cell and environment causing breaks

colinhanrahan:
Current repo:
- `Cells` are indexed by `(x, y)` or `(width, height)`.
- `Environment` is indexed by `[height][width]`. See line 5 of `environment.py`.
- Due to this mismatch, any environment size that is not square breaks.
- "North" is considered to be in the positive y direction (downwards on GUI). 

This PR:
- Having to choose one of the two coordinate systems, sticking with `Cell` `(x, y)` makes more sense.
- `Environment` is now indexed by `[width][height]` (required changes to multiple files).
- "North" is now in the negative y direction (upwards on GUI, which is more accurate). Affects season placement and neighbor naming.
- Standardizes `find*Neighbor()` functions to look the same as `findNorthNeighbor()`.
- Removes unused variables in sugar and spice peak generation.

### DIFF
--- a/cell.py
+++ b/cell.py
@@ -72,31 +72,19 @@ class Cell:
         return neighborWealth
 
     def findEastNeighbor(self):
-        eastWrapAround = self.environment.width
-        eastIndex = self.x + 1
-        if eastIndex >= eastWrapAround:
-            eastIndex = 0
-        eastNeighbor = self.environment.findCell(eastIndex, self.y)
+        eastNeighbor = self.environment.findCell((self.x + 1 + self.environment.width) % self.environment.width, self.y)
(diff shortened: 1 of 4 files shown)


nkremerh:
Going to add testing and validating this as an exercise for the first week of research work. Stay tuned.

nkremerh:
Confirmed nearly identical trends across 50 seeds out to 1k timesteps.

colinhanrahan:
Note: determinism with the current repo is maintained without the north-south switch. The only difference with the north-south switch is that the order of neighbor cells is slightly different, so it will still be slightly different after shuffles.

PR #65, opened 2024-06-15
Fixes #64 agent blindness with tribes and one resource

### LINKED ISSUE #64: findWelfare() incorrectly returns 0 when environment only has one resource and tribes are turned on
I found this during testing for `cultural_tagging.json`. If there is only one resource, the other resource's `*MetabolismProportion` will be 0, so that resource will multipy into `welfare` as `0 ** 0` (just 1). However, when tribes are included, the `*TagPreferences` value will be _nonzero_, so the updated welfare will include 0 to a nonzero power. This makes `findWelfare()` return 0 for all cells, so agents move in a random direction close to them with no useful information about the resources they are gathering. In practice, tribal agents are blind with functional range 1.

As a side note, tribe-based welfare is added near the end of the book (page 124), after most of the examples. Even if we fix this bug, all of our example simulations that include tribes will be adversely affected by this feature. Welfare is modified in a random and often unhelpful way by an agent's tribe, so they are forced to make bad decisions. Fixing this would require adding a config option.

colinhanrahan:
This behavior is particularly evident in `cultural_tagging.json`. I could also add the config option to this PR if we decide on that. See #64.


nkremerh:
Go ahead and add the configuration option for this one. Let's call it `agentTagPreferences` where it's a Boolean value. True indicates to use tag-based (tribe-based) preferences for resources. False indicates no tag-based preferences are used.

If I'm reading the code correctly, this should be reflected in line 1034 in `findWelfare` and as a new attribute for the agent.

colinhanrahan:
@nkremerh It doesn't really fit in with the code we have for `randomizeAgentEndowments` — there's no precedent for Boolean variables per agent. It will be the same for every agent and never change during the course of the simulation, so maybe it's closer to a `Sugarscape` variable like `seed` or `debug`, than an `Agent` one? We can still name it `agentTagPreferences`, but it just won't be part of each agent's config.

colinhanrahan:
I added the new option to the "true default" config in `sugarscape.py`. This means it should be automatically set to false for all of the examples, which is what we want. However, this does bring up the issue that the formatting of example configs will go out-of-date every time we add a config option. I think we should only include the options we are changing from the default in the example configs and allow `sugarscape.py` to handle the rest of them. This would keep the configs cleaner and more up-to-date. I'll add it to our Kanban backlog as a potential change.

nkremerh:
Correct, the tag preferences will work just like movement mode, etc.

Yes, ideally each example will only include options which deviate from the default config at the bottom of `sugarscape.py`.

agent.py
```diff
@@ -1031,13 +1033,11 @@ def findWelfare(self, sugarReward, spiceReward):
...
-        if self.tags != None and len(self.tags) > 0:
+        if self.tagPreferences == True and self.tags != None and len(self.tags) > 0:
             # Tribe could have changed since last timestep, so recheck
             self.tribe = self.findTribe()
             fractionZeroesInTags = self.tagZeroes / len(self.tags)
             fractionOnesInTags = 1 - fractionZeroesInTags
-            spiceMetabolism = spiceMetabolism if spiceMetabolism > 0 else 1
-            sugarMetabolism = sugarMetabolism if sugarMetabolism > 0 else 1
             tagPreferences = (sugarMetabolism * fractionZeroesInTags) + (spiceMetabolism * fractionOnesInTags)
             if tagPreferences == 0:
```
nkremerh, line 1042: If removing lines 1039-1040, then this condition should be set to `<=` instead to prevent a negative preference since it's used in exponentiation.

nkremerh (changes requested):

agent.py
colinhanrahan: Good catch!

nkremerh (approved):

nkremerh:
Final change: add a description of the new config option to the README.

colinhanrahan:
How does this look?

PR #66, opened 2024-06-15
Fixes #56 agent range is not what is expected for other agents and at the end of timestep

### LINKED ISSUE #56: Agents seem to have fewer cells in range than their vision should let them see (including disease effects)
After digging into `ethics.py` to change out some hardcoded values, I noticed that agents are frequently having fewer cells found in the `cellsInRange` attribute (which should be equivalent to the number of cells found by a call to `findVision()` multiplied by the kind of neighborhood).

Using the default von Neumann neighborhood scheme, an agent should have `agent.findVision() * 4` cells available to consider. There are cases where a healthy agent (so no disease effects) is seeing `(agent.findVision() - 1) * 4` cells. This is not universal, which is puzzling. This led me to check for disease, but healthy agents are experiencing this behavior. It should also be noted that not _all_ healthy agents experience this, either.

This is apparent in seed `12345` using the rest of the default configuration options (recommended to ratchet down the timesteps to 200 or fewer and set the decision model to `benthamHalfLookaheadBinary`).

colinhanrahan:
#56
- Updates agent initialization to find `cellsInRange` and `neighborhood`, matching child creation
- Agents now update `cellsInRange` after they move to correctly reflect range after being cured by disease and in other agent's ethical calculations

Just for future reference, this will definitely increase runtime.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -47,7 +47,9 @@ class Agent:
 
         self.alive = True
         self.age = 0
+        self.cellsInRange = []
         self.lastMoved = -1
+        self.neighborhood = []
         self.vonNeumannNeighbors = {"north": None, "south": None, "east": None, "west": None}
         self.mooreNeighbors = {"north": None, "northeast": None, "northwest": None, "south": None, "southeast": None, "southwest": None, "east": None, "west": None}
         self.socialNetwork = {"father": None, "mother": None, "children": [], "friends": [], "creditors": [], "debtors": []}
@@ -83,8 +85,6 @@ class Agent:
         self.visionModifier = 0
         self.aggressionFactorModifier = 0
         self.fertilityFactorModifier = 0
-        self.cellsInRange = self.findCellsInRange()
-        self.neighborhood = self.findNeighborhood()
 
     def addChildToCell(self, mate, cell, childConfiguration):
         sugarscape = self.cell.environment.sugarscape
--- a/agent.py
+++ b/agent.py
@@ -47,9 +47,7 @@ class Agent:
 
         self.alive = True
         self.age = 0
-        self.cellsInRange = []
         self.lastMoved = -1
-        self.neighborhood = []
         self.vonNeumannNeighbors = {"north": None, "south": None, "east": None, "west": None}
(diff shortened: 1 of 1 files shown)


nkremerh:
What if I told you this actually cut runtime in half?

I ran the same 100 seeds out to 500 timesteps (for ease of data collection) for the head of the current master branch and this branch using the `benthamHalfLookahead` decision model. Both branches produced 100% better outcomes from their starting populations (this number can be obtained by running `make plots`).

This branch, however, produced 10% lower population on average (1,000 for the master branch vs. 900 for this branch). Every other metric tracked was more or less the same.

Time for master branch:
```
real    13m56.042s
user   263m53.385s
sys       0m9.615s
```

Time for this branch:
```
real     7m49.831s
user   138m14.411s
sys       0m7.189s
```

Thoughts? Population size definitely contributes the most to runtime. A 10% difference in population leading to a nearly 50% reduction in runtime seems a bit extreme.

colinhanrahan:
Comparing the two branches using Snakeviz on the same seed (and using `benthamHalfLookaheadBinary`), it seems like almost every function is getting its runtime halved. That's pretty odd. Maybe the 10% difference is actually causing that big of a difference? I'll investigate this more.

colinhanrahan:
> * Updates agent initialization to find `cellsInRange` and `neighborhood`, matching child creation

This addition is redundant with `sugarscape.py` lines 144-146. The `sugarscape.py` lines should be kept and my addition will be removed because all agents need to be created first in order for neighborhoods to be correct on the first timestep.

colinhanrahan:
I ran a test on both branches where I ended runs based on the number of agents (I ended at 500) instead of the number of timesteps. This PR is more time intensive as expected. If you want to mass test this, modify line 604 in `sugarscape.py`:
```python
- while t <= timesteps and len(self.agents) > 0:
+ while t <= timesteps and len(self.agents) > 0 and len(self.agents) <= 500:
```

colinhanrahan:
Note: one more `findCellsInRange()` should be added to `catchDisease()` for agents who already had their timestep but are catching a disease from another agent. I'll add it to the Kanban board.

PR #68, opened 2024-06-16
Code style change

amuller26:
Replaced all instances of strings using .format() with the f"{variable}" (I think I created this correctly?)


colinhanrahan (changes requested):
- It looks like this branch is a few commits behind the repo, which is to be expected with all the changes we're making. To resolve conflicts, you can `fetch` recent changes from the repo and then `merge` them into this branch. I can send you an email with more detail on how to do this if you want.
- There are a few more strings using `.format()` in `data/run.py` and `plots/plot.py`.

PR #70, opened 2024-06-18
Adds pollution timers and fixes pollution diffusion logic

colinhanrahan:
**Pollution diffusion logic change**
- Current repo: the average of the current cell and its neighbors becomes the new pollution for all of those cells. Cells are looped through in normal nested for loop fashion and changed one-by-one, resulting in a sort of top-left bias.
- Book/this PR: the average of the current cell's neighbors (_not_ including the current cell) becomes the new pollution for _only_ the current cell. All calculations are done in parallel and then applied to all cells at once, preventing directional bias. See page 48 for comparison.

**Pollution timers**
- Adds a config option that decides when agent consumption and production pollution should be turned on
- Adds a config option that decides when pollution diffusion should be turned on

TODO for this PR:
- [x] Add a timer config option to turn pollution _off_ for page 127. Diffusion does not have an "off" timer in the book but one can be added.
- [x] Add correct timers to configs for other examples that use pollution.
- [x] Primarily, fix the behavior seen in `pollution_formation.json`.

These two features can be split into separate PRs if necessary, but they kind of need to work together. I'm trying to replicate `pollution_formation.json` and running into some issues, which is why I'm opening this PR now. It seems like agents produce pollution behind themselves which encourages them to move in straight lines (kind of reminiscent of Conway's Game of Life). In the example, most agents also die before they get the chance to move back to the sugar peaks, which makes sense to me because they have to survive 50 timesteps with barely any food. The book's diagram makes less sense to me. Can somebody try this out for themselves and help brainstorm solutions to these discrepancies? See pages 47-49.


nkremerh:
Both these features should be in this single PR. While the _exact_ behavior of the diagrams from the book might not be entirely reproducible, this gets us closer to what the book intended. That is still a win.

nkremerh:
I think I found a promising lead: I think I implemented the pollution ratio incorrectly (lines 606-607 of `agent.py`).

When trying to mesh all the features together (something not really done in the book), I originally interpreted the resources to pollution ratio as being the calculated agent welfare compared to pollution. However, I think the welfare calculation needs to incorporate the resource to pollution ratio for both sugar and spice inside it.

Line 606 becomes:
`welfare = self.findWelfare(((cell.sugar + welfarePreySugar) / (1 + cell.pollution)), ((cell.spice + welfarePreySpice) / (1 + cell.pollution)))`

Line 607 becomes:
`cellWealth = welfare # This line can be refactored away`

With this change, a society using the default decision model (greedy) with no combat or disease manages to treat low pollution as a preference whereas before it seemed to treat it as a necessity. However, the starting population needs to be much higher than our standard (1250 instead of 250) for society to stick around longterm.

Does this change get you closer to a resolution?

colinhanrahan:
> With this change, a society using the default decision model (greedy) with no combat or disease manages to treat cells having low pollution as a preference whereas before it seemed to treat it as a necessity. However, the starting population needs to be much higher than our standard (1250 instead of 250) for society to stick around longterm.

I think this make sense. When pollution was tested long-term in the book (p. 128), there was a delay before it was turned on to allow society to start off on the right foot. If agents start polluting right away, they will scatter, which means they will be much less likely to reproduce.

`pollution_formation` does seem to match better with the book after this change. In pictures 3-4 on page 49, agents are traveling to cells with no sugar at all, which doesn't make sense considering that _sugar / (pollution + 1)_ = 0 when _sugar_ is 0. Our agents never leave the radius of the sugar peak, which makes sense because they are still primarily looking for sugar. 

There are some other small differences. After pollution diffuses everywhere, our agents generally remain on the very edge of sugar peak tiers, which makes sense because higher tiers produce more consumption pollution which is cascading down to the surroundings. We could add some margin of error to try to achieve more "natural" behavior by rounding pollution instead of allowing it to be infinitely precise, but other than that, I think the example is good. I'll finish up the rest of the features on the TODO.

colinhanrahan:
Take a look at the behavior in `trade_pollution.json` before you merge.

nkremerh:
Add in verification that start and stop timesteps are in the right order, if the max is greater than the configured `timesteps` for the run, if the max is -1 set the value to be equal to `timesteps`, and if the min is below 0 set it to 0.

colinhanrahan:
@nkremerh Are the edge cases handled how you want?

README
```diff
@@ -320,9 +320,13 @@ environmentMaxTribes: int
     Set the maximum number of tribes in the starting population.
     Default: 0
 
-environmentPollutionDiffusionDelay: int
```
nkremerh, line 323: This option should not be removed since it does something different than the added features. It sets the timestep interval for when diffusion should happen (e.g. every 10 timesteps).

colinhanrahan: You're right, that was an accident. I meant to remove the ones that I added.

PR #71, opened 2024-06-18
Adds network overlay for book visualizations

colinhanrahan:


gui.py
```diff
@@ -231,12 +268,93 @@ def doTimestep(self):
...
+
     def doWindowClose(self, *args):
         # Indicate to per-timestep rendering to stop rendering
         self.stopSimulation = True
         self.window.destroy()
         self.sugarscape.toggleEnd()
 
+    def drawLines(self):
+        lineCoordinates = set()
+        if self.activeNetwork.get() == "Neighbors (Von Neumann)":
```
nkremerh, line 291: Can this be extended to cover both neighborhood modes?

If the neighborhood mode is von Neumann, show the von Neumann neighborhood. Otherwise, the Moore neighborhood.

colinhanrahan: Sure, added.

PR #72, opened 2024-06-18
Adds variable quadrant sizing

colinhanrahan:
- New config setting: `"environmentQuadrantSizeFactor"` between 0 and 1.
- The default is 1, which means the quadrants agents start in each take up exactly one quarter of the environment.
- The combat examples from the book use 20x20 quadrants = a quadrant size factor of 0.8, so their configs have been modified to account for this.
- `findActiveQuadrants()` was wasting runtime being called every timestep, so it is now only called once at the beginning of the simulation.
- The list returned by `findActiveQuadrants()` is now a list of lists with each sublist being a quadrant. This will be very useful for adding specific tribes to quadrants, which can be added after this PR goes through.

### DIFF
--- a/README
+++ b/README
@@ -319,6 +319,10 @@ environmentPollutionDiffusionDelay: int
     Set the delay interval in timesteps when pollution is diffused across the environment.
     Default: 0
 
+environmentQuadrantSizeFactor: float
+    Set the proportion of each corner of the screen taken up by the agents' starting quadrants.
+    Default: 1
+
 environmentSeasonalGrowbackDelay: int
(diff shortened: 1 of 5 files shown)


nkremerh:
Changing the `findActiveQuadrants` behavior does not have an impact on examples that have the agent replacement rule enabled, does it?

colinhanrahan:
Agent replacements spawn in whatever quadrants you enable as active. So if you only use quadrants 2 and 4 for the starting agents, the replacements will also only spawn in 2 and 4. I'm pretty sure this is the intended behavior.

PR #73, opened 2024-06-18
Replaces strings using .format() with f"{variable}"

amuller26:


CREDITS
```diff
@@ -7,9 +7,11 @@ Colin Hanrahan
 Willem Hueffed
 Nate Kremer-Herman
 Maria Milkowski
+Anna Muller
 Joshua Palicka
 Mariana Shuman
 Lucas Vorkoper
+random
```
colinhanrahan, line 14: I don't think "random" is supposed to be here.

plots/plot.py
```diff
@@ -98,26 +98,24 @@ def printHelp():
...
     columnHeads += '\n'
     file.write(columnHeads)
 
     for i in range(totalTimesteps + 1):
         line = str(i)
         for model in dataset:
-            line += " {0} {1} {2} {3} {4} {5}".format(dataset[model]["meanMetrics"]["population"][i], dataset[model]["meanMetrics"]["agentMeanTimeToLive"][i],
-                                                  dataset[model]["meanMetrics"]["agentStarvationDeaths"][i], dataset[model]["meanMetrics"]["agentCombatDeaths"][i],
-                                                  dataset[model]["meanMetrics"]["agentWealthTotal"][i], dataset[model]["meanMetrics"]["meanAgeAtDeath"][i])
+            line += f" {dataset[model]["meanMetrics"]["population"][i]} {dataset[model]["meanMetrics"]["agentMeanTimeToLive"][i]} {dataset[model]["meanMetrics"]["agentStarvationDeaths"][i]} {dataset[model]["meanMetrics"]["agentCombatDeaths"][i]} {dataset[model]["meanMetrics"]["agentWealthTotal"][i]} {dataset[model]["meanMetrics"]["meanAgeAtDeath"][i]}"
```
colinhanrahan, line 118: I think we should still split this line up for readability, maybe something like this:
```python
line += f" {dataset[model]["meanMetrics"]["population"][i]} {dataset[model]["meanMetrics"]["agentMeanTimeToLive"][i]}"
line += f" {dataset[model]["meanMetrics"]["agentStarvationDeaths"][i]} {dataset[model]["meanMetrics"]["agentCombatDeaths"][i]}"
line += f" {dataset[model]["meanMetrics"]["agentWealthTotal"][i]} {dataset[model]["meanMetrics"]["meanAgeAtDeath"][i]}"

```

colinhanrahan (changes requested):
I have 2 small suggestions, otherwise looks good!

CREDITS
amuller26: That was a test that I forgot to delete.

nkremerh:
Looks like there is some version history management issue going on (some commits are out of place in the timeline). Follow up with me individually when you have time.

nkremerh:
After productive discussion with @amuller26, she will be creating a new PR with the history in the correct order. The joys of learning Git.

PR #74, opened 2024-06-18
Adds config option to disable wraparound

colinhanrahan:
While working on this, I overhauled agents' Von Neumann and Moore neighbors.
- Removed `agent.vonNeumannNeighbors` and `agent.mooreNeighbors`
- Removed `update*Neighbors()` functions
- Added `agent.socialNeighbors` because we still need asymmetrical adjacency neighborhoods for network overlay
- All other functions that used adjacent neighbors now use `agent.cell.neighbors.values()`
- `cell.neighbors`, previously a list, is now a dictionary with directions. If wraparound is turned off, cells on the edge will just not have directions that go out of bounds. For example, cells on the north edge will not have `"northwest"`, `"north"`, or `"northeast"` in their `cell.neighbors` dictionary.

Possible changes:
- Add another dictionary to distinguish between a cell's Moore and Von Neumann neighbors. `cell.neighbors` uses whichever neighborhood mode you picked in the config.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -51,6 +51,7 @@ class Agent:
         self.cellsInRange = []
         self.lastMoved = -1
         self.neighborhood = []
+        self.socialNeighbors = []
         self.socialNetwork = {"father": None, "mother": None, "children": [], "friends": [], "creditors": [], "debtors": []}
         self.diseases = []
         self.fertile = False
@@ -220,6 +221,7 @@ class Agent:
 
         # Keep only debtors and children in social network to handle outstanding loans
         self.socialNetwork = {"debtors": self.socialNetwork["debtors"], "children": self.socialNetwork["children"]}
+        self.socialNeighbors = []
         self.neighborhood = []
         self.diseases = []
 
@@ -431,7 +433,7 @@ class Agent:
             self.lastSpice = self.spice
             self.lastMoved = self.timestep
             self.moveToBestCell()
-            self.updateSocialNetwork()
+            self.updateNeighbors()
             self.collectResourcesAtCell()
             self.doUniversalIncome()
             self.doMetabolism()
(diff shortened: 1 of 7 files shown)


colinhanrahan:
The new logic in `updateNeighbors()` looks just like `cell.findNeighborAgents()`. That function is only called in `doLending()`, which could just use `cell.neighbors` and add a check for `agent != None` like everything else. We can probably remove `findNeighborAgents()` (or if you really want, change the logic of everything else to use this function).

agent.py
```diff
@@ -51,8 +51,7 @@ def __init__(self, agentID, birthday, cell, configuration):
         self.cellsInRange = []
         self.lastMoved = -1
         self.neighborhood = []
-        self.vonNeumannNeighbors = {"north": None, "south": None, "east": None, "west": None}
-        self.mooreNeighbors = {"north": None, "northeast": None, "northwest": None, "south": None, "southeast": None, "southwest": None, "east": None, "west": None}
+        self.socialNeighbors = []
```
nkremerh, line 54: Let's just call this `self.neighbors`.

agent.py
```diff
@@ -1319,13 +1300,15 @@ def updateMeanIncome(self, sugarIncome, spiceIncome):
         self.spiceMeanIncome = (alpha * spiceIncome) + ((1 - alpha) * self.spiceMeanIncome)
 
     def updateNeighbors(self):
-        self.updateVonNeumannNeighbors()
-        self.updateMooreNeighbors()
+        self.socialNeighbors = []
+        for neighborCell in self.cell.neighbors.values():
```
nkremerh, line 1304: Is there a clean way to do this as a list comprehension one-liner?

agent.py
nkremerh: Especially since we have the `updateNeighbors` method for agents.

nkremerh (changes requested):

agent.py
colinhanrahan: Sure!

agent.py
colinhanrahan: Renamed in new commit.

nkremerh (approved):

PR #78, opened 2024-06-19
Fixes use of magic numbers in hedonic calculus

nkremerh:
Modifies hedonic calculus to use agent properties rather than magic numbers
Adds agent finding their new cells in range when catching disease
Adds agent lookahead discount when considering future rewards in hedonic calculus
Fixes GUI getting proper vision, movement, and metabolism values for agent printouts

### DIFF
--- a/README
+++ b/README
@@ -149,7 +149,8 @@ agentLoanDuration: [int, int]
     Default: [0, 0]
 
 agentLookaheadDiscount: [float, float]
-    Set the agent's discount when considering possible rewards in future timesteps.
+    Set the agent's discount applied to considering possible rewards in future timesteps.
+    Note: This feature is used in agent decision models besides the "none" and "rawSugarscape" models.
     Default: [0, 0]
 
 agentLookaheadFactor: [int, int]
--- a/README
+++ b/README
@@ -148,6 +148,10 @@ agentLoanDuration: [int, int]
     Set the agent's provided loan duration in timesteps.
     Default: [0, 0]
 
+agentLookaheadDiscount: [float, float]
+    Set the agent's discount when considering possible rewards in future timesteps.
+    Default: [0, 0]
+
 agentLookaheadFactor: [int, int]
     Set the agent's consideration of metabolic costs in timesteps ahead.
     Default: [0, 0]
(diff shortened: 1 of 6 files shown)


nkremerh:
@colinhanrahan please double-check my work.

Nearly all changes (once bug fixes included are also applied) match exactly with the outcomes before this PR.

Adding a separate agent lookahead discount property causes a deviation. The way initial agents are provided endowments and how children are provided endowments is set up such that the random number generator is shielded from changes when we add new agent properties. I'm likely overlooking something small.

Please also feel free to suggest optimizations.

colinhanrahan:
- For agent lookahead discount—a deviation in that it breaks from determinism but outcomes are similar, or a deviation in that it causes a significant change in behavior?
- You might want to consider further distinguishing `agentLookaheadDiscount` from `agentLookaheadFactor` in the README by including something about how it only applies to ethical agents. They sound pretty similar as of now.
- Make sure the new `futureIntensity` calculation works as intended when wraparound is turned off. Cells on the edges with, for example, 2 neighbors will be treated differently than other cells with 4 neigbors, which might not be what you want.

nkremerh:
Breaks determinism but outcomes are practically the same. The seed I was testing led to deviation at timestep 20, which is odd since that is after agents reproduce. So, the determinism change isn't caused by children being born. I'll take another dive into it.

Adding a note to the README.

Future intensity calculation verified.

colinhanrahan:
For reference, my seed broke determinism on timestep 93.

nkremerh:
Fixed. I placed it in the paired endowments (could be temporary, but it makes sense to put it there permanently).

colinhanrahan:
Interesting—I didn't find anything that would change random states for `parentEndowments`.

As far as optimizations go, I was expecting for it to be slower given the new `len` calls in `ethics.py`, but it looks fine.

colinhanrahan:
@nkremerh Alright, I found the issue by comparing the `parentEndowments` and `pairedEndowments` loops:

```python
        for endowment in parentEndowments:
-           index = random.randrange(2)
            random.seed(self.childEndowmentHashes[endowment] + self.timestep)
+           index = random.randrange(2)
            endowmentValue = parentEndowments[endowment][index]
            childEndowment[endowment] = endowmentValue
```

The `random.randrange()` should be called _after_ the seed is set, not before. This fixes determinism on the seeds I tested. We should add this so we don't run into the same issue when adding new agent config options in the future.

PR #81, opened 2024-06-19
Adds config option for number of diseases per starting agent

colinhanrahan:
This is a new kind of config option we haven't seen before. It's not passed down to children _or_ used for replacement agents. (Options like the starting quadrant aren't passed down to children but are used for replacement agents.) To account for this, I had to add an agent variable that's only used for the initial population of agents. There might be a better way to do this by separating the logic for this new endowments from the rest of the endowments, and then using that in `configureDiseases` directly so you don't have to go through the agent.

Edit: writing a new `randomizeStartingAgentEndowments` method for endowments that are only used for the initial population (this is the only one so far).


nkremerh:
You're right this is wonky. This is more of an environmental/simulation level option, so I'd like to avoid making this a property of the first generation agents.

You know how many agents there will be, and you know the range of starting diseases. You could create a list containing the number of diseases to give each agent (with however many repeated occurrences from cycling through the range over and over). Shuffle the list. Then divvy out the number of diseases by popping the list and applying that many diseases to an agent. All this can be done in `sugarscape.py` without changing the agent data structure.

colinhanrahan:
@nkremerh This looks a little better. I decided not to write a new function to randomize endowments that are only for initial agents since it's a pretty niche use case.

nkremerh:
It would be preferable to maintain the original way of seeding initial diseases in the simulation. If a number of starting diseases is set, but diseases per agent is `[0, 0]`, then the current master branch behavior should happen (infect X number of agents with a unique disease at the start).

colinhanrahan:
The book specifies X random diseases will be given to each agent, not X random diseases that they are not immune to, so I left that in with the new logic. We should probably just go one way or the other. I tested
- making sure each disease has different tags (currently, they can have the same tags)
- making sure agents were not immune to the diseases they were being given (with the new distribution logic)

and both of these resulted very quickly in population extinction.

agent.py
```diff
@@ -229,19 +229,17 @@ def doDeath(self, causeOfDeath):
 
     def doDisease(self):
         diseases = self.diseases
+        random.shuffle(diseases)
         for diseaseRecord in diseases:
             diseaseTags = diseaseRecord["disease"].tags
             start = diseaseRecord["startIndex"]
```
nkremerh, line 235: Since `end` is removed, may as well also remove `start`.

sugarscape.py
```diff
@@ -154,24 +154,35 @@ def configureDiseases(self, numDiseases):
...
+                        diseases.remove(newDisease)
+                        break
+            if len(diseases) > 0 and ("all" in self.debug or "sugarscape" in self.debug):
+                print(f"Could not place {len(diseases)} diseases.")
+        else:
+            minStartingDiseases = startingDiseases[0]
+            maxStartingDiseases = startingDiseases[1]
+            currStartingDiseases = minStartingDiseases
+            for agent in self.agents:
+                agentDiseases = random.sample(self.diseases, currStartingDiseases)
```
nkremerh, line 180: Does this ensure a roughly equal spread of _each_ disease created for the population, or will this result in a semi-random distribution of the diseases present? Is it possible (though unlikely) that a single disease could remain unplaced?

sugarscape.py
```diff
@@ -154,24 +154,35 @@ def configureDiseases(self, numDiseases):
...
+                for newDisease in agentDiseases:
                     agent.catchDisease(newDisease)
-                    self.diseases.append(newDisease)
-                    diseases.remove(newDisease)
-                    break
-        if len(diseases) > 0 and ("all" in self.debug or "sugarscape" in self.debug):
```
nkremerh, line 173: Some form of error message needs to remain for both cases ([0,0] for starting diseases and the new base case) to warn users if their desired number of diseases aren't able to be placed.

nkremerh (changes requested):

sugarscape.py
colinhanrahan: The book specifies a completely random sample from the master disease list. It is possible that a single disease could remain unplaced in some configs, but it's extremely unlikely for the book examples since we have so many agents and 4/10 or 10/25 disease are given to each agent. I could modify this logic to use the pseudorandom uniform distribution we use for other endowments if you want.

sugarscape.py
colinhanrahan: The new logic gives every agent a certain number of diseases, so the only issue would be that the program is attempting to give more unique diseases to an agent than exist in the master list. This would be a better fit for `verifyConfiguration`, so I'll add it there.

sugarscape.py
nkremerh: Yes, let's keep it consistent with the others. I think the random sample approach by the book is not worth reproducing exactly.

colinhanrahan:
- Do we need `sugarscape.diseases`, or can I remove it? It's not used for anything.
- `disease_basic.json` _immediately_ goes extinct with the Hamming distance check, so that's something to think about.

nkremerh:
We may as well keep it. The list of diseases could be useful for future research.

I take it you mean the sheer number of diseases and the provided immune system length make survival impossible?

colinhanrahan:
Yes. I'm checking the 10 out of 25 diseases per agent version, which is supposed to be more difficult to survive, but not that much more difficult. To be fair, the actual symptoms diseases inflict are _never disclosed_ in the book other than a vague "metabolic fees." I'll reduce the metabolism penalties and try to get a closer match with the book.

agent.py
```diff
@@ -229,19 +229,16 @@ def doDeath(self, causeOfDeath):
 
     def doDisease(self):
         diseases = self.diseases
+        random.shuffle(diseases)
         for diseaseRecord in diseases:
             diseaseTags = diseaseRecord["disease"].tags
-            start = diseaseRecord["startIndex"]
-            end = diseaseRecord["endIndex"] + 1
             immuneResponse = [self.immuneSystem[i] for i in range(diseaseRecord["startIndex"], diseaseRecord["endIndex"] + 1)]
```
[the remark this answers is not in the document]
colinhanrahan: Done. Hamming distance isn't really defined for strings of unequal lengths. We just iterate over the immune system and truncate any part of the disease that's outside of it to the right. When the disease is longer than the immune system, we're not checking if the right side of the disease tags might result in a closer Hamming distance if we truncate the left side instead. Perhaps it would be more interesting to slide whichever string is smaller over the window of the larger one? That way, if the disease is longer, you're checking all possible positions rather than only ones where the disease starts at immune system[0] or greater. I was working on this for a bit but it's big enough to be a separate PR. What are your thoughts?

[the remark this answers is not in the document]
nkremerh: If I'm understanding it correctly, it shouldn't matter if we find a more optimal match as a substring. So long as the strings don't match, the disease never cures. In the case of a disease tag being longer than an immune system, finding a perfectly matching substring will have no effect on the resulting behavior.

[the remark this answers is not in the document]
colinhanrahan: I guess that's true—I was thinking more about the theory of asymmetrical Hamming distance than the practice. To be fair, the specific tag being flipped might have an impact on other diseases the agent has.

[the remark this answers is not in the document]
nkremerh: Good point from a computational perspective, but the immune systems don't have to be optimal much like real immune systems encountering a novel infection; they may be absolutely powerless to stop it.

nkremerh:
Something seems broken now: making the single change to the default config `startingDiseasesPerAgent` to `[0,1]` causes 100% of agents to have a disease in my testing.

nkremerh (approved):

colinhanrahan:
This works with everything I tested but having three conditionals per loop isn't particularly glamorous.

PR #83, opened 2024-06-20
Memoizes range for efficiency

colinhanrahan:
This breaks determinism, but only because the order of cells in agents' `cellsInRange` changes. I'm looking into why, because the order should really be the same. Also, a good next step to optimize `findNeighborhood` would be to make `cellsInRange` a set and find its intersection with `Sugarscape`'s `self.agents` list.

If the memoization takes up too much storage, we can instead store the difference between the current range and the previous range (the additional cells added) rather than storing the whole list for every distance.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -676,15 +676,14 @@ class Agent:
         return bestFriend
 
     def findCellsInRange(self, newCell=None):
-        cell = self.cell if newCell == None else newCell
         vision = self.findVision()
         movement = self.findMovement()
         cellRange = min(vision, movement)
         if cellRange > 0:
             if (self.visionMode == "cardinal" and cellRange == vision) or (self.movementMode == "cardinal" and cellRange == movement):
-                allCells = self.cell.environment.findCellsInCardinalRange(cell.x, cell.y, cellRange)
+                allCells = self.cell.cardinalRanges[cellRange]
             else:
-                allCells = self.cell.environment.findCellsInRadialRange(cell.x, cell.y, cellRange)
+                allCells = self.cell.radialRanges[cellRange]
             if newCell == None:
                 self.cellsInRange = allCells
             return allCells
--- a/agent.py
+++ b/agent.py
@@ -680,7 +680,7 @@ class Agent:
         cellRange = min(vision, movement)
         allCells = {}
         for i in range(1, cellRange + 1):
-            allCells = {**allCells, **cell.ranges[i]}
+            allCells = allCells.update(cell.ranges[i])
         if newCell == None:
             self.cellsInRange = allCells
         return allCells
--- a/agent.py
+++ b/agent.py
@@ -680,7 +680,7 @@ class Agent:
         cellRange = min(vision, movement)
         allCells = {}
         for i in range(1, cellRange + 1):
-            allCells = allCells.update(cell.ranges[i])
+            allCells.update(cell.ranges[i])
         if newCell == None:
             self.cellsInRange = allCells
         return allCells
--- a/agent.py
+++ b/agent.py
@@ -680,15 +680,12 @@ class Agent:
         vision = self.findVision()
         movement = self.findMovement()
         cellRange = min(vision, movement)
-        if cellRange > 0:
-            if (self.visionMode == "cardinal" and cellRange == vision) or (self.movementMode == "cardinal" and cellRange == movement):
-                allCells = cell.cardinalRanges[cellRange]
-            else:
-                allCells = cell.radialRanges[cellRange]
-            if newCell == None:
-                self.cellsInRange = allCells
-            return allCells
-        return []
+        if cellRange <= 0:
+            return []
+        allCells = cell.ranges[cellRange]
+        if newCell == None:
+            self.cellsInRange = allCells
+        return allCells
 
     def findChildEndowment(self, mate):
         parentEndowments = {
--- a/agent.py
+++ b/agent.py
@@ -682,7 +682,9 @@ class Agent:
         cellRange = min(vision, movement)
         if cellRange <= 0:
             return []
-        allCells = cell.ranges[cellRange][:]
+        allCells = []
+        for i in range(1, cellRange + 1):
+            allCells.extend(cell.ranges[i])
         if newCell == None:
             self.cellsInRange = allCells
         return allCells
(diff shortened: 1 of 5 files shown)


colinhanrahan:
The order was different because `self.cellsInRange` was a reference to `cell.ranges[distance]` instead of a copy, which was then shuffling `cell.ranges[distance]` in `findBestCell`. This has been fixed and determinism is now maintained with the current repo.

environment.py
```diff
@@ -85,6 +85,19 @@ def findCellNeighbors(self):
             for j in range(self.height):
                 self.grid[i][j].findNeighbors(self.neighborhoodMode)
 
+    def findCellRanges(self):
+        minDistance = min(self.sugarscape.configuration["agentVision"][0], self.sugarscape.configuration["agentMovement"][0])
+        maxDistance = min(self.sugarscape.configuration["agentVision"][1], self.sugarscape.configuration["agentMovement"][1])
```
nkremerh, line 90: I sent you off on an incorrect approach. Setting a min and max before runtime won't account for diseases which inhibit (or expand) vision/movement. Perhaps we should take a _boil the ocean_ approach and go from `0...max(environmentHeight, environmentWidth)` or something similar. We compute the distance from each cell to every other cell.

I suppose this can be represented as a table in an `Environment` object. Any object (cell or agent) needing to look up distance can consult the table. That sounds like an appropriately ugly yet fast solution. Perfect for memoization.

nkremerh: Finding all cells within a given range would take a constant number of list operations (grabbing each cell up to some distance away in all relevant directions based on vision/movement mode).

colinhanrahan: We _could_ have a table in `Environment`, but since the ranges are already tied to the cells, I don't see why they shouldn't be a `Cell` variable. I added logic to account for disease penalties.

colinhanrahan:
Finding 50 different radial ranges for 50x50 cells in the environment takes exponentially too long. Thoughts?

nkremerh:
If cell X can see cell Y at range 10, then cell Y can also see cell X at range 10, etc. See if that provides a useful reduction?

colinhanrahan:
Ready for testing.

nkremerh:
Initial testing (50 seeds out to 2k timesteps) shows a significant time reduction:

Current repo head:
```
real	15m38.914s
user	231m6.808s
sys	0m10.446s
```

This branch:
```
real	11m14.852s
user	166m25.475s
sys	0m7.853s
```

Will test more thoroughly before merge, but I will be checking style and general fit and finish first.

agent.py
```diff
@@ -898,8 +895,8 @@ def findNeighborhood(self, newCell=None):
         else:
             newNeighborhood = self.findCellsInRange(newCell)
         neighborhood = []
-        for neighborCell in newNeighborhood:
-            neighbor = neighborCell["cell"].agent
+        for neighborCell in newNeighborhood.keys():
```
nkremerh, line 898: I'm okay with using `keys()` and `items()` going forward, but there are a lot of instances where I set things up to be more old school (and less Pythonic). As a separate PR (once this is merged), someone should walk through the code and replace instances of loops which can be refactored to use `keys()` or `items()`.

nkremerh:
Let me know if any of the readability changes I made accidentally broke any logic.

colinhanrahan:
No, everything looks good. In terms of other optimizations, I'm eyeing `findNeighborhood` to see if I can use set intersection with an agent's cells in range and a set of all cells with agents on them. I'm not sure if it would actually be faster with larger numbers of agents.

As for this PR, there might be a way to speed up radial range calculations when the max distance is smaller. Currently, even when the max distance is less than the environment size, every cell combination has to be checked (2500 choose 2). I'm going to keep thinking about this for now, but barring any further breaks or issues, the current implementation can be merged.

environment.py
```diff
@@ -85,59 +113,48 @@ def findCellNeighbors(self):
...
+        for i in range(numCells):
+            x1, y1 = cellCoords[i]
+            for j in range(i + 1, numCells):
+                x2, y2 = cellCoords[j]
+                deltaX = self.findWraparoundDistance(x1 - x2, self.width)
+                deltaY = self.findWraparoundDistance(y1 - y2, self.height)
+                if deltaX > maxDeltaX or deltaY > maxDeltaY:
+                    continue
+                deltaPair = tuple(sorted((deltaX, deltaY)))
+                distance = distanceTable[deltaPair]
```
nkremerh, line 147: There's a runtime error for this when setting vision and movement modes to `radial`, neighborhood mode to `moore`, and keeping the rest of the default config the same (exception: using the `bentham` model for survivability):

```
python sugarscape.py --conf config.json
Traceback (most recent call last):
  File "/home/user/sugarscape/sugarscape.py", line 1047, in <module>
    S = Sugarscape(configuration)
        ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/colinscape/sugarscape.py", line 42, in __init__
    self.configureEnvironment(configuration["environmentMaxSugar"], configuration["environmentMaxSpice"], configuration["environmentSugarPeaks"], configuration["environmentSpicePeaks"])
  File "/home/user/colinscape/sugarscape.py", line 200, in configureEnvironment
    self.environment.findCellRanges()
  File "/home/user/colinscape/environment.py", line 131, in findCellRanges
    self.findRadialCellRanges(maxDeltaX, maxDeltaY, maxDeltaRadius, cellCoords)
  File "/home/user/colinscape/environment.py", line 147, in findRadialCellRanges
    distance = distanceTable[deltaPair]
               ~~~~~~~~~~~~~^^^^^^^^^^^
KeyError: (0, 25)
```

colinhanrahan: Should be fixed. Wraparound calculations should always use the environment borders even if the max deltas are smaller than the borders.

nkremerh:
Testing 20 seeds with `radial` vision and movement and `moore` neighborhoods:

Current repo head:
```
real	60m26.204s
user	229m56.582s
sys	0m2.806s
```

This branch:
```
real	8m40.028s
user	75m46.485s
sys	0m5.386s
```

Even with the same seeds and configs, the outcomes between the two branches are different due to when random events happen. However, the societal outcomes are more or less the same.

PR #86, opened 2024-06-23
Add config option to disable cultural transmission

amuller26:
Added the toggle and adjusted agents and children accordingly. I am not sure if this is the expected outcome. In the simulation, it seems that children who are born can be born to a new tag, but I am not sure where and how it is happening.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -40,7 +40,7 @@ class Agent:
         self.sugar = configuration["sugar"]
         self.sugarMetabolism = configuration["sugarMetabolism"]
         self.tags = configuration["tags"]
-        self.tagging = configuration["tags"]
+        self.tagging = configuration["tagging"]
         self.tagPreferences = configuration["tagPreferences"]
         self.tradeFactor = configuration["tradeFactor"]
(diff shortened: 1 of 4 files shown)


colinhanrahan:
Some things to consider:
- add the new config option to the README
- set `agentTagging` to true for most of the examples (but I think one or two of them do use rule K)

Otherwise looks good! I like that you kept the tagging logic self-contained in `doTagging` rather than expanding it out to `doTimestep`.

agent.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, agentID, birthday, cell, configuration):
         self.sugar = configuration["sugar"]
         self.sugarMetabolism = configuration["sugarMetabolism"]
         self.tags = configuration["tags"]
+        self.tagging = configuration["tags"]
```
colinhanrahan, line 43: This should be `configuration["tagging"]` or else `doTagging` will happen even if tagging is set to false.

colinhanrahan (changes requested):

agent.py
amuller26: Done.

PR #89, opened 2024-06-27
Adds agent decision model tribal factor

colinhanrahan:
`agentDecisionModelTribalFactor` is currently a paired endowment (with other decision model-related traits), but it could be a regular endowment instead.

0 will make agents disregard their own tribe and 1 will make them disregard other tribes, like `selfishnessFactor`.

### DIFF
--- a/ethics.py
+++ b/ethics.py
@@ -17,7 +17,6 @@ class Bentham(agent.Agent):
         cellNeighborWealth = cell.findNeighborWealth()
         globalMaxWealth = cell.environment.globalMaxSugar + cell.environment.globalMaxSpice
         cellValue = 0
-        selfishnessFactor = self.selfishnessFactor
         neighborhoodSize = len(self.neighborhood)
         futureNeighborhoodSize = len(self.findNeighborhood(cell))
         for neighbor in self.neighborhood:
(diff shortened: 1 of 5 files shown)


ethics.py
```diff
@@ -70,18 +69,25 @@ def findEthicalValueOfCell(self, cell):
...
                 else:
                     neighborCellValue = neighbor.decisionModelFactor * ((certainty * proximity) * ((extent * (intensity + duration)) + (discount * (futureExtent * (futureIntensity + futureDuration)))))
-            if selfishnessFactor > 0:
+
+            if self.decisionModelTribalFactor >= 0:
+                if neighbor.findTribe() == self.findTribe():
+                    neighborCellValue *= self.decisionModelTribalFactor
+                else:
+                    neighborCellValue *= 1 - self.decisionModelTribalFactor
+            if self.selfishnessFactor > 0:
```
colinhanrahan, line 78: By the way, this should be `>= 0` for Altruist models.

nkremerh: Very true! Patch it in.

nkremerh:
@amuller26 Review, please.

amuller26:
In `sugarscape.py` at line 975, I think it needs to be 
```
if configuration["agentDecisionModelTribalFactor"][1] > 1:
    configuration["agentDecisionModelTribalFactor"][1] = 1
elif configuration["agentDecisionModelTribalFactor"][0] < 0:
    configuration["agentDecisionModelTribalFactor"][0] = -1
```
It was throwing TypeError originally.

colinhanrahan:
> In `sugarscape.py` at line 975, I think it needs to be
> 
> ```
> if configuration["agentDecisionModelTribalFactor"][1] > 1:
>     configuration["agentDecisionModelTribalFactor"][1] = 1
> elif configuration["agentDecisionModelTribalFactor"][0] < 0:
>     configuration["agentDecisionModelTribalFactor"][0] = -1
> ```
> 
> It was throwing TypeError originally.

Good catch, thanks! I also added a `sort()`.

amuller26 (approved):

sugarscape.py
```diff
@@ -970,6 +972,12 @@ def verifyConfiguration(configuration):
     if configuration["agentTagStringLength"] > 0 and configuration["environmentMaxTribes"] > configuration["agentTagStringLength"]:
         configuration["environmentMaxTribes"] = configuration["agentTagStringLength"]
 
+    configuration["agentDecisionModelTribalFactor"].sort()
```
nkremerh, line 975: We should do a sort of every config option that represents a range of values near the top of `verifyConfiguration`. @amuller26, create a Kanban card for that and add that to your plate.

colinhanrahan: Just make sure you keep in mind the config options that can have `[positive number, -1]` like the pollution timeframes.

nkremerh: Exactly. There are edge cases to keep in mind.

PR #90, opened 2024-06-27
Modifies loops to use .items()

amuller26:
Also simplified three loops in agent.py

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -247,10 +247,9 @@ class Agent:
         diseaseCount = len(self.diseases)
         if diseaseCount == 0:
             return
-        neighborCells = self.cell.neighbors.values()
         neighbors = []
-        for neighborCell in neighborCells:
-            neighbor = neighborCell.agent
+        for cell in self.cell.neighbors.values():
+            neighbor = cell.agent
             if neighbor != None and neighbor.isAlive() == True:
                 neighbors.append(neighbor)
         random.shuffle(neighbors)
@@ -470,10 +469,9 @@ class Agent:
         self.sugarPrice = 0
         self.spicePrice = 0
         self.findMarginalRateOfSubstitution()
-        neighborCells = self.cell.neighbors.values()
         traders = []
-        for neighborCell in neighborCells:
-            neighbor = neighborCell.agent
+        for cell in self.cell.neighbors.values():
+            neighbor = cell.agent
             if neighbor != None and neighbor.isAlive() == True:
                 neighborMRS = neighbor.marginalRateOfSubstitution
                 if neighborMRS != self.marginalRateOfSubstitution:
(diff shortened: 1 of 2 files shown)


nkremerh:
@colinhanrahan Review, please.

sugarscape.py
```diff
@@ -546,12 +546,12 @@ def randomizeAgentEndowments(self, numAgents):
             sexDistributionCountdown = math.floor(sexDistributionCountdown / (maleToFemaleRatio + 1)) * maleToFemaleRatio
 
         for i in range(numAgents):
-            for config in configurations:
-                configurations[config]["endowments"].append(configurations[config]["curr"])
-                configurations[config]["curr"] += configurations[config]["inc"]
-                configurations[config]["curr"] = round(configurations[config]["curr"], configurations[config]["decimals"])
-                if configurations[config]["curr"] > configurations[config]["max"]:
-                    configurations[config]["curr"] = configurations[config]["min"]
+            for config in configurations.items():
```
colinhanrahan, line 549: I think `.values()` is a better fit here. With `.items()`, you're setting `config` to a tuple of the key (the name of the config option) and the value (the sub-dictionary with `"endowments"` etc.). `.values()` will let you access just the values, which in this case is the sub-dictionary you're currently accessing with `config[1]`. So it should look something like `for config in configurations.values(): config["endowments"].append(...)`.

colinhanrahan (changes requested):

agent.py
```diff
@@ -822,10 +820,9 @@ def findTimeToLive(self, ageLimited=False):
 
     def findEmptyNeighborCells(self):
         emptyCells = []
-        neighborCells = self.cell.neighbors.values()
-        for neighborCell in neighborCells:
-            if neighborCell.agent == None:
-                emptyCells.append(neighborCell)
+        for cell in  self.cell.neighbors.values():
```
nkremerh, line 823: There's an extra space after `in` on this line.

nkremerh:
This should a separate branch from your master as you will get a merge conflict when I do a squash and merge of this PR. Your master branch's history will have a few commits which will soon be squashed into one. Closing this and will wait for a new one to open from a separate branch.

agent.py
```diff
@@ -247,10 +247,9 @@ def doDisease(self):
         diseaseCount = len(self.diseases)
         if diseaseCount == 0:
             return
-        neighborCells = self.cell.neighbors.values()
         neighbors = []
-        for neighborCell in neighborCells:
-            neighbor = neighborCell.agent
+        for cell in self.cell.neighbors.values():
```
nkremerh, line 251: On second thought, this change should be reverted. It doesn't gain us anything and incurs an extra function call without added clarity for what's happening.

agent.py
nkremerh, line 823: On second thought, this change should be reverted. It doesn't gain us anything and incurs an extra function call without added clarity for what's happening.

PR #91, opened 2024-06-28
Adds live histograms and Lorenz curve

colinhanrahan:
I just wanted to get this up as a draft to show that I have, indeed, been making progress. I'm keeping the graph stats separate from runtime stats for now but they can be merged later if you want to log them.

TODO:
- [x] Add Lorenz curve
- [x] Handle edge cases (simulation end etc.)
- [x] Add axes and numbers
- [x] General cleanup (+ remove hardcoded values)
- [x] Add option to save frames/animations


nkremerh:
Don't worry about saving frames. If a user wants to plot the data after they run the simulation, they should be able to do so from the log. Each timestep, make sure any new (aggregate) data needed for the charts is added to the runtime stats. Let's consider these to be similar to the social network graphs. They're ephemeral.

If the value needed for a chart is simply taking two (or more) values already present in the runtime stats, don't worry about adding that computed value redundantly into the log unless it's a tricky equation.

nkremerh:
A design decision I'll make here is to treat these graphs as another tab in the simulation window (like the social network graphs) rather than a popup window. If a user wants to look back at the visualization, they can switch tabs back.

colinhanrahan:
Things I figure you'll want:
- [ ] Graph titles and axis labels
- [ ] Full dynamic resizing?
- [ ] Live Gini coefficient on Lorenz curve as seen in book

nkremerh:
Yes, add in resizing.

nkremerh:
If the window is resized before opening a graph tab, the interface resets to the default size and jumps to the origin (upper left corner) of the screen.

colinhanrahan:
> If the window is resized before opening a graph tab, the interface resets to the default size and jumps to the origin (upper left corner) of the screen.

Are you saying that even without interacting with the graph menu, the window jumps to the top left corner when you try to resize it? That's not happening on my Macbook, so your window manager must be doing something different. The issue is probably that the `.geometry` calls don't have a set x and y (only width and height), but it'll be tricky for me to debug since it's not an issue on my computer.

nkremerh:
Not quite. This behavior occurs when the window loads, I resize it, then I click on one of the graphs (say, Gini coefficient).

colinhanrahan:
The resize when switching from environment to graphs is intentional (and should occur whether or not you resize the window beforehand). The default size/aspect ratio for the environment should not necessarily be the same as the default size/aspect ratio for graphs, and I picked a size for `initialGraphWidth` and `initialGraphHeight` that I thought looked good. I can revert this if you want. However, the jumping to top left behavior is not intentional nor does it happen on my computer. I'm not sure how to fix this, but I'll push a few changes and then you can see if it behaves any better.

nkremerh:
It works! Bug is fixed.

gui.py
```diff
@@ -145,6 +162,9 @@ def configureEnvironment(self):
 
     def configureEnvironmentColorNames(self):
         return ["Pollution"]
+
+    def configureGraphNames(self):
+        return ["Sugar Histogram", "Spice Histogram", "Tag Histogram", "Age Histogram", "Lorenz Curve"]
```
nkremerh, line 167: Let's call `Lorenz Curve` the `Gini Coefficient` graph instead.

Please also alphabetize these and the social network graph options.

colinhanrahan: Done, but I kept some backend references to `lorenzCurve` for clarity (the polygon tag, for example).

PR #92, opened 2024-06-29
Prevents inherited curses creating zombie agents

colinhanrahan:
When an agent dies from starvation, their sugar or spice can be negative. Attempting to pass on this negative wealth to children or other heirs results in the heirs _losing_ wealth instead of gaining wealth. If heirs have already completed their timestep but this curse brings one of their resources into the negative, they will be removed from `Sugarscape`'s `self.agents` by `removeDeadAgents` without going through the proper `doDeath` process. This results in zombie agents that are kept alive by their cells referencing them and eventually overwritten when a new agent enters the cell.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -261,6 +261,11 @@ class Agent:
     def doInheritance(self):
         if self.inheritancePolicy == "none":
             return
+        # Prevent negative inheritance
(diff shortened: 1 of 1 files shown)


nkremerh:
Good catch!

Issue #46, opened 2024-03-15
Dataset parsing and graph creation does not scale with experimental complexity


nkremerh:
We are beginning to use the software to investigate research questions which require more complex experimentation. The current software pipeline allows for easily investigating the impact of different decision models across six key metrics across time.

However, what if we wanted to gather data on a combination of `[list of decision models] + [list of metabolisms] + [list of trade factors]`, etc? The pipeline cannot accommodate that without herculean changes to `plots/plot.py`. Further, what if we wanted to present data not across time but across different variables (such as `tradeFactor` or `selfishnessFactor`)? Again, the plotting script would require radical revisions.

Combine new data collection combinations _and_ new ways to present them, and we end up with a spaghetti code mess. What's the preferable option? Provide a querying interface for the dataset and a middleware layer which can take the output of queries and plot them as graphs. The goal would be to use _strictly_ base installation Python modules and software which is regularly available on most Linux distributions. We want to minimize our dependency footprint.

Python has a provided `sqlite3` module, and SQLite does support JSON. There's a query language. This is a broader conversation, but we are already rapidly approaching a reasonable upper bound on complexity for our plotting script before it becomes unreasonable to expect new users to contribute to it in the future.

colinhanrahan:
I haven't looked into `sqlite3` yet, but I _was_ running into the issue of not enough customization for data collection when working on my multiple decision models branch.

I'd like to propose a new structure for data collection where `config.json` splits into 2 files:
1. `config.json`
- for normal, single simulation use
- stores `"sugarscapeOptions"` only, does not store `"dataCollectionOptions"`
2. `data_collection_config.json`
- built for data collection
- stores `"dataCollectionOptions"` at top of file
- `"dataCollectionOptions"` now stores a new `"pathToDefaultConfig"` variable which, by default, is set to `"config.json"`
- below that, stores individual JSON objects representing simulations. You only include the settings you want to change; the rest will automatically be set to the settings from `"pathToDefaultConfig"`
- there will be 3 JSON objects below by default, one for each simulation of the current default setup for data collection we have now. But new JSON objects can be added underneath to add more simulations for data collection.

Here's a short example:
```
{
    "__README__": "Default values for data collection. Details can be found in the README.",
    "dataCollectionOptions": {
        "bashAlias": "bash",
        "jobUpdateFrequency": 5,
        "numParallelSimJobs": 1,
        "numSeeds": 100,
        "pathToDefaultConfig": "config.json",
[...]

PR #95, opened 2024-06-30
Adds matplotlib as plotting engine

nkremerh:
Adds matplotlib dataset plotting
Removes gnuplot dataset plotting
Modifies configuration options and documentation to match new plotting mechanism

With seemingly few new adoptees of gnuplot, I've made the decision to move onto a new plotting engine. While hesitant to incorporate additional Python modules into the codebase, the plotting being an optional component makes this a minor ding against my goal of keeping everything contained in the base Python install. An alternative to this approach would be to use GNU Octave (which is compatible with Matlab).

### DIFF
--- a/plots/plot.py
+++ b/plots/plot.py
@@ -1,5 +1,7 @@
 import getopt
 import json
+import matplotlib.pyplot
+import matplotlib.ticker
 import os
 import re
 import sys
@@ -33,9 +35,9 @@ def parseDataset(path, dataset, totalTimesteps):
                     continue
                 if entry not in datacols:
                     datacols.append(entry)
-                if entry not in dataset[model]["meanMetrics"]:
-                    dataset[model]["meanMetrics"][entry] = [0 for j in range(totalTimesteps + 1)]
-                dataset[model]["meanMetrics"][entry][i-1] += item[entry]
+                if entry not in dataset[model]["metrics"]:
+                    dataset[model]["metrics"][entry] = [0 for j in range(totalTimesteps + 1)]
+                dataset[model]["metrics"][entry][i-1] += item[entry]
             i += 1
         if rawJson[-1]["population"] == 0:
             dataset[model]["died"] += 1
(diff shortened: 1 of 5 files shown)


plots/plot.py
```diff
@@ -103,178 +106,50 @@ def printSummaryStats(dataset):
...
+    matplotlib.pyplot.rcParams["font.family"] = "serif"
+    matplotlib.pyplot.rcParams["font.serif"] = ["Times New Roman"]
+    matplotlib.pyplot.rcParams["font.size"] = 20
+    figure, axes = matplotlib.pyplot.subplots()
+    axes.set(xlabel = "Timestep", ylabel = label, xlim = [0, totalTimesteps])
+    x = [i for i in range(totalTimesteps + 1)]
+    lines = []
+    colors = ["magenta", "cyan", "gold", "black"]
+    c = 0
+    modelStrings = {"bentham": "Utilitarian", "egoist": "Egoist", "altruist": "Altruist", "none": "Raw Sugarscape", "rawSugarscape": "Raw Sugarscape"}
```
colinhanrahan, line 139: We are allowing multiple decision models per simulation in our data collection config but not handling the naming correctly. This line along with line 142 throws `KeyError` when multiple decision models are in the same simulation.

colinhanrahan (changes requested):

plots/plot.py
colinhanrahan: For line 142, you can check if the model is a list, and use `' '.join(modelStrings[model] for model in modelList)` if it is. This will make the names pretty long. My solution to this in my old multiple decision models branch was to abbreviate, e.g. `benthamHalfLookaheadBinary` -> `bHLB`.

nkremerh: This PR is a drop-in replacement for the old plotting script, which also did not accurately plot mixed decision models or the long name ones. This PR now fixes that bug while also sidestepping this issue for now. The plotting script has always been a more bespoke tool for our publications. That's the primary goal for #46: make this a general purpose tool for users to query their dataset and produce plots. This is a significant step toward that goal.

PR #97, opened 2024-07-09
Adds experimentation for tribal assimilation

amuller26:
I added the experimentation as an alternative to runSimulation(). I added a boolean to the configuration to determine if the user wants to run the tribal assimilation analysis. It also just prints out the result in the command line.

### DIFF
--- a/analysis.py
+++ b/analysis.py
@@ -1,6 +1,8 @@
 import csv
+import json
 import sys
 
+# Will only produce results for the first run in a CSV file
 def analyzeTribalAssimilation(log):
     tribeSummary = {}
     numTribes = 0
@@ -19,6 +21,17 @@ def isTribeDominant(numTribes, percent):
         return True
     return False
 
+def loadCSVFile(fileName):
+    logFile = open(fileName)
+    log = csv.DictReader(logFile)
+    return log
+
+# Will only parse if there is only one run logged in the file
+def loadJSONFile(filename):
+    logFile = open(fileName)
+    log = json.load(logFile)
+    return log
+
 def parseFile(filename):
     split = filename.split(".")
     format = split[-1]
@@ -35,7 +48,13 @@ if __name__ == "__main__":
     log = None
 
     if fileFormat == "csv":
-        logFile = open(fileName)
-        log = csv.DictReader(logFile)
+        log = loadCSVFile(fileName)
+    elif fileFormat == "json":
+        log = loadJSONFile(fileName)
+    else:
+        print("Unsupported file format.")
+        exit(1)
 
     analyzeTribalAssimilation(log)
+
+    exit(0)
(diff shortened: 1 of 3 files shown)


nkremerh:
A conceptual pivot: we should be able to track _all_ of the emergent behaviors we want to quantify at our fingertips for _any_ run of the simulation. We should be including new metrics in the log that will allow us to do some analysis after-the-fact and determine when/how/why emergent behaviors happened.

amuller26:
I added analysis.py so it runs right after runSimulation()

nkremerh:
That's okay for a first draft. Let's build something that works first then make it look good later!

sugarscape.py
```diff
@@ -331,6 +332,17 @@ def generateTribeTags(self, tribe):
         random.shuffle(tags)
         return tags
 
+    def getTribeCount(self):
```
colinhanrahan, line 335: Rather than creating a list and then converting it to a dictionary, why not just use one data structure the whole time? Either a list or dictionary would be fine, but a dictionary might be better just because it makes the relationship between tribe number and count more explicit. You can keep the `"Tribe"` in the key if you want, but you could also just print it when needed in the `__str__`.

How does this look to you?
```python
    tribeCounts = {f"Tribe {i}": 0 for i in range(numTribes)}
    for agent in self.agents:
        tribeCounts[agent.tribe] += 1
    return tribeCounts
```

Alternatively, following my other comment on `tribeCounts`:
```python
    tribeCounts = [0 for i in range(numTribes)]
    for agent in self.agents:
        tribeCounts[agent.tribe] += 1
    return tribeCounts
```

analysis.py
```diff
@@ -0,0 +1,44 @@
+import json
+class Analysis:
+    def __init__(self, config):
+        self.log = self.loadJSON(config["logfile"])
```
colinhanrahan, line 4: Reading the logfile causes a break if there is more than one simulation log in the file. This could be handled by reading only the last simulation log.

analysis.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def getTribeKeys(self):
+        tribeKeys = []
+        keysList = list(self.log[0].keys())
+        for key in keysList:
+            if "Tribe" in key:
+                tribeKeys.append(key)
+        return tribeKeys
+
+    def isTribeDominant(self, tribe, totalPopulation):
```
colinhanrahan, line 31: The assimilation flag activates within the first few timesteps because we're checking if any tribe has more than an equal proportion. Even if there is 1 more agent in a tribe than expected (ex. 49:51 agents), the tribe will be considered dominant. I know we're still trying to figure out what metric to use for assimilation—I just wanted to give an update on the current metric.

amuller26:
Do you want me to add a Makefile option to run? I don't know how to do that when choosing which log file to use.

nkremerh:
Kernel of functionality added in 20718d1.

PR #98, opened 2024-07-12
Fixes Gini coefficient

colinhanrahan:
The main problems were:
- Using `agent.wealth` (see my email)
- If `area += (height - wealth) / 2` is supposed to calculate the area _under_ the Lorenz curve, it should just be `area += height`. If it's supposed to calculate the area between the line of equality and the Lorenz curve, then it doesn't work because the `giniCoefficient =` line treats it as the area under the Lorenz curve (among other reasons).

I can rework it to still calculate the area under the line of equality if you don't like the magic number 0.5, but it seems intuitive to normalize it.

### DIFF
--- a/sugarscape.py
+++ b/sugarscape.py
@@ -683,8 +683,9 @@ class Sugarscape:
             cumulativeWealth += wealth
             lorenzCurveArea += cumulativeWealth / totalWealth
         lorenzCurveArea /= len(agentWealths)
-
-        giniCoefficient = round((0.5 - lorenzCurveArea) / 0.5, 2)
+        # When normalized, the triangle under the equality line has base and height 1
+        equalityLineArea = 0.5
+        giniCoefficient = round((equalityLineArea - lorenzCurveArea) / equalityLineArea, 3)
         return giniCoefficient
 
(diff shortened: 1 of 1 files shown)


sugarscape.py
```diff
@@ -674,15 +674,17 @@ def toggleRun(self):
...
         for wealth in agentWealths:
-            height += wealth
-            area += (height - wealth) / 2
-        lineOfEquality = (height * len(agentWealths)) / 2
-        giniCoefficient = round((lineOfEquality - area) / max(1, lineOfEquality), 2)
+            cumulativeWealth += wealth
+            lorenzCurveArea += cumulativeWealth / totalWealth
+        lorenzCurveArea /= len(agentWealths)
+
+        giniCoefficient = round((0.5 - lorenzCurveArea) / 0.5, 2)
```
nkremerh, line 687: We can keep the magic number since the two areas we care about add up to 0.5. Put that in a variable called `areaUnderLineOfEquality` or similarly descriptive variable and add a _brief_ comment explaining the value.

Looking at the book, let's also round to 3 rather than 2 decimal places.

PR #105, opened 2024-07-16
Adds missing examples from book

colinhanrahan:


colinhanrahan:
@nkremerh I'm feeling a bit stuck on this PR.
pgs. 64-65: Added, but they do not resemble the book.
pgs. 88-89: Decided not to add until diagonal migration waves is figured out.
pgs. 91-93: Added, but they do not resemble the book.

Can you take a look at pgs. 118-119, pg. 121, and pgs. 124-127 and confirm that we should add these? For example, 118-119 are just `trade_basic.json` but with different vision ranges.

nkremerh:
I'm okay with not including simple rule variants in the examples.

How far do the examples included here deviate from the book? It's okay if the patterns or emergent behaviors are of different magnitude.

colinhanrahan:
The oscillation examples never develop into a consistent oscillation. The example for page 93 does not develop into culturally segregated communities, and it is very unlikely that it ever would given that agents from both tribes migrate to both peaks.

nkremerh:
Let's cut the proto-history example. The rest should be good to go (even if they don't exactly match the book patterns).

colinhanrahan:
Ready for further review.

PR #106, opened 2024-07-19
Adds movement, vision, metabolism coloring (pgs. 59-62)

colinhanrahan:
Thoughts:
- Like I did for network colors, I hardcoded these colors because they're only used in a conditional which explains their purpose and they match the book exactly. However, they can be moved to `self.colors` if needed.
- Instead of following the book—with blue being used for low values and red being used for high values—we could use a gradient with similar logic to `recolorByResourceAmount`. Green to blue to purple might be a good option compared to the colors used in the environment.
- Added instance variables because I didn't want to spend resources calculating the means per agent per timestep.


nkremerh:
This is great!

I like the gradient idea. Perhaps we go from red (worst performing) to green (best performing) or similar? This becomes tricky since the best metabolism is actually the min of the configuration whereas the best vision and movement are both their respective max value in the config.

colinhanrahan:
> I like the gradient idea. Perhaps we go from red (worst performing) to green (best performing) or similar?

A red to green gradient means you'll either have a weird brown color in between or you use yellow, which looks the same as sugar. That's why I was trying to figure out better colors to use.

> This becomes tricky since the best metabolism is actually the min of the configuration whereas the best vision and movement are both their respective max value in the config.

I don't see this as an issue. The book just acknowledges the color difference and in the diagrams the "better" metabolism is blue while the "better" vision is red.

colinhanrahan:
Actually, it's looking pretty good! One particular thing I discovered was that we were calculating interpolated colors for every non-agent cell for every timestep. I know from previous runtime analysis that this was taking up runtime. I've updated it to use a lookup table (since I needed to refactor `recolorByResourceAmount` anyway).

I'm ignoring disease penalties for now because agents can technically have dozens of diseases and it would mess up the color range for disease-free agents. That said, we could still account for disease-penaltied values if they are within the default color range, which I currently haven't done. It would technically be slightly more accurate, but since we can't fully account for disease penalties, it might be better to just keep the colors disease-free.

nkremerh:
Feel free to mess with the colors to not overlap with environmental colorings. Keep in mind pollution is purple, so the original concept of including purple would also not do well in all situations.

Even though the book does not resolve the issue of lower metabolism being good while higher vision is good, that doesn't mean we should follow suit. If we're using gradient shading and using the same colors across all three properties, then the best values possible should all be shaded the same. Likewise with the worst values, regardless whether that is a low or high number.

Let's ignore diseases since what we want to demonstrate is the evolution of certain genes over time. Adding in disease will conflate things. Good intuition.

colinhanrahan:
Green and red looks good so far. Flipped metabolism to match the other metrics—green is better.

colinhanrahan:
@nkremerh On the topic of color gradients, I took a stab at making the color blending between sugar and spice peaks more pleasant. Here's what I came up with:
Screenshot 2024-07-21 at 10 09 32 PM
Versus the current blending:
Screenshot 2024-07-23 at 2 03 16 PM
Thoughts? Not that it matters much.

nkremerh:
I like it. The stark contrast of the tan center is helpful for explaining that the area is a resource-rich, preferable zone for agents. If the gradient you have here comes about a quarterway or halfway further toward the current coloration, that would be ideal.

colinhanrahan:
@nkremerh 
Screenshot 2024-07-23 at 4 12 51 PM
Screenshot 2024-07-23 at 4 13 05 PM
These are the best I can come up with for now. No worries if they're not what you're looking for.

nkremerh:
Seeing the options, go ahead and open a PR for the first gradient you did. I like it most out of the options.

PR #107, opened 2024-07-20
implements and logs carrying capacity

hadiyachishti:
implements and logs carrying capacity

### DIFF
--- a/carryingCap.py
+++ b/carryingCap.py
@@ -1,0 +1,56 @@
+import json
+import random
+import matplotlib.pyplot as plt
+from sugarscape import Sugarscape, verifyConfiguration
+
+def runSimulationAndLog(configuration):
+    sim = Sugarscape(configuration)
+    
+    populationOverTime = []
+
+    while not sim.end:
+        sim.doTimestep()
+        populationOverTime.append(sim.runtimeStats["population"])
+        
+        # Check for a stable population indicating carrying capacity
+        if len(populationOverTime) > 10 and all(populationOverTime[-10:]) == populationOverTime[-1]:
+            break
+    
+    with open(f"population_log_{configuration['seed']}.json", 'w') as f:
+        json.dump(populationOverTime, f, indent=4)
+    
+    finalPopulation = sim.runtimeStats["population"]
+    return finalPopulation
+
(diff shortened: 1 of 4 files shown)


carryingCap.py
```diff
@@ -0,0 +1,56 @@
...
+    with open(configFile, 'r') as f:
+        baseConfig = json.load(f)
+    baseConfig = baseConfig.get("sugarscapeOptions", baseConfig)
+
+    carryingCapacityReached = 0
+    extinctionOccurred = 0
+
+    for _ in range(numRuns):
+        config = baseConfig.copy()
+        config["seed"] = random.randint(0, 10000)
```
colinhanrahan, line 35: For extra security, we use `random.randint(0, sys.maxsize)` in other areas of the repo.

carryingCapPlot.py
```diff
@@ -0,0 +1,38 @@
...
+                populationLogs.append(json.load(f))
+
+    return populationLogs
+
+def plotPopulationOverTime(populationLogs):
+    carryingCapacities = []
+    
+    for log in populationLogs:
+        plt.plot(log, label=f'Seed {populationLogs.index(log) + 1}')
+        carryingCapacities.append(max(log))  # assuming the carrying capacity is the max population reached
```
colinhanrahan, line 20: In `runSimulationAndLog` you consider the carrying capacity to be roughly the last ten values of the list. Maybe you can use the same logic here for runs that initially overshoot their carrying capacity?

sugarscape.py
```diff
@@ -50,6 +50,11 @@ def __init__(self, configuration):
         self.activeQuadrants = self.findActiveQuadrants()
         self.configureAgents(configuration["startingAgents"])
         self.configureDiseases(configuration["startingDiseases"])
+        if configuration["headlessMode"] == False:
```
colinhanrahan, line 53: Looks like this is redundant with line 58 below?

nkremerh:
This is ready for another look. @hadiyachishti please email me what the kernel here is for how we calculate carrying capacity. I think that belongs in `sugarscape.py` and should be part of updating the runtime stats (like Gini coefficient).

nkremerh:
Resolved by 846d0a2.

PR #111, opened 2024-07-25
Adds diagonal waves example

colinhanrahan:
Diagonal migration is generated when the sugar in "front" of an agent is greater than all the sugar it can see behind it, even accounting for growback, for a continuous period of time. With the combination of max vision 10, max sugar 4, and the shape of our sugar peaks, the small situational window that generates the emergent behavior is nearly lost. Our example might be slightly closer to the book (using max sugar 4) if we copied the sugar distribution exactly. Hervé Lange's diagonal migration example uses a max sugar of 10 and our repo generates a similar result with those settings.

@nkremerh I'll send you some related videos via Outlook.

### DIFF
--- a/examples/diagonal_waves.json
+++ b/examples/diagonal_waves.json
@@ -1,0 +1,12 @@
+{
+    "__README__": "Emergent agent migration waves with constant sugar growback (pgs. 42-43 of Growing Artificial Societies).",
+    "agentMovement": [1, 1000],
+    "agentVision": [1, 10],
+    "agentStartingSugar": [5, 25],
+    "environmentMaxSugar": 10,
+    "environmentQuadrantSizeFactor": 0.8,
+    "environmentStartingQuadrants": [4],
+    "seed": 12345,
+    "startingAgents": 300,
+    "timesteps": 500
+}


nkremerh:
Fantastic! Is there a means of perpetually spawning waves? Currently, once the second peak becomes colonized, the behavior shifts to the more traditional clusters of agents around the peaks.

Perhaps have a low max age and agent replacement? I know that deviates from the book, so please let me know if you have any other ideas.

colinhanrahan:
That may or may not work—I'll look into it. The problem there is that diagonal waves are not generated by the agent locations, but by directional differences in the sugar distribution. These directional differences are initially created by agents being forced to move out of the corner in a wave shape and continue while the agents migrate. Once agents migrate the first time, the sugar distribution becomes more noisy, so the same kind of wave might not happen again.

The other option is the second video I sent you: fill the entire sugarscape with sugar and make it regenerate more slowly so that agents are forced to stay on the move. There are clear, permanent waves of agents here.

Another thought: maybe we could put two peaks in the far corners of a larger environment with a small sugar bridge between them and seasons turned on. When the season changes, agents would have to cross the bridge, which might create a wave. But they would have to be really packed together when the season changes because the forced movement based on locations they cannot move (cells with other agents) creates the wave shape.

nkremerh:
Remind me: is the growback rate constant or immediate for this example?

colinhanrahan:
Growback is constant—if it was immediate, agents would have no incentive to continue moving in the same direction.

nkremerh:
That's what I figured. It's a puzzle. If this is as close as we get given the simulation setup, that's okay.

colinhanrahan:
I'm not seeing any more waves with agent replacement than without. Are you interested in adding something like video #2 I sent you as an example?

nkremerh:
What you have here is sufficient for demonstrating the behavior.

PR #112, opened 2024-07-26
Fixes bug where too many replacement agents are created

colinhanrahan:
One of my configs with 300 initial and 300 replacement agents was jumping up to 492 agents. `totalCells` is the number of empty spawning locations, not the total number of spawn locations.

I moved agent replacements to before the GUI draw so that the number of agents on screen and the GUI agent counter are always equal to the replacement agents (if 300, there will always be 300 agents on screen). You might prefer that agents are replaced after the GUI draw (the current behavior) so that you still get a sense of how many agents died last timestep (# replacements - agent counter on GUI). In that case, the other changes—removing the extra GUI draw and correcting the number of configured agents—should still stay.

I'm also thinking that we should `updateRuntimeStats` once after creating the GUI and before running any `sugarscape.doTimestep` so that the counters on the menu bar are not all incorrectly 0 at timestep 0. They are correct if logging is enabled because we `updateRuntimeStats` at the beginning of `startLog`, but they should be updated even if there is no log.

### DIFF
--- a/sugarscape.py
+++ b/sugarscape.py
@@ -45,6 +45,8 @@ class Sugarscape:
         self.debug = configuration["debugMode"]
         self.keepAlive = configuration["keepAlivePostExtinction"]
         self.agents = []
+        self.agentsReplaced = 0
+        self.agentsBorn = 0
         self.deadAgents = []
         self.diseases = []
         self.activeQuadrants = self.findActiveQuadrants()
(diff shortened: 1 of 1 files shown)


nkremerh:
Correctness is preferred, but we can achieve that by adding in another runtime metrics (`agentsReplaced` and `agentsBorn` for reproduction). That way none of these calculations have to occur in a special order anymore.

colinhanrahan:
Is this what you were looking for? I removed the `agentReproduced` because we don't use and it makes no sense to me.

PR #115, opened 2024-07-29
Adds metrics for diseases in a separate log file

amuller26:
Adds metrics for all diseases logged individually. Disease stats are stored in a separate log file, which can be either a CSV file or JSON file.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -170,6 +170,9 @@ class Agent:
         caughtDisease = {"disease": disease, "startIndex": startIndex, "endIndex": endIndex}
         if infector != None:
             caughtDisease["infector"] = infector
+            if infector not in disease.infectors:
+                disease.infectors.append(infector.ID)
+        disease.infected += 1
         self.diseases.append(caughtDisease)
         self.updateDiseaseEffects(disease)
         self.findCellsInRange()
--- a/config.json
+++ b/config.json
@@ -87,6 +87,7 @@
         "keepAlivePostExtinction": false,
         "logfile": "log.json",
         "logfileFormat": "json",
+        "logfileDisease": "diseaselog.json",
         "neighborhoodMode": "vonNeumann",
         "profileMode": false,
         "screenshots": false,
--- a/disease.py
+++ b/disease.py
@@ -12,6 +12,12 @@ class Disease:
         self.aggressionPenalty = configuration["aggressionPenalty"]
         self.tags = configuration["tags"]
         self.configuration = configuration
+        self.infectors = []
+        self.infected = 0
+
+    def resetRStats(self):
+        self.infectors = []
+        self.infected = 0
 
     def __str__(self):
         return f"{self.ID}"
--- a/sugarscape.py
+++ b/sugarscape.py
@@ -65,6 +65,9 @@ class Sugarscape:
                            "maxSugar": 0, "maxSpice": 0, "maxWealth": 0}
         self.log = open(configuration["logfile"], 'a') if configuration["logfile"] != None else None
         self.logFormat = configuration["logfileFormat"]
+        self.diseaselog = open(configuration["logfileDisease"], 'w')
+        self.diseasesStats = []
+        self.diseasesStats.append({"disease": None, "timestep": None, "totalInfected": None, "newlyInfected": None, "infectors": None, "r": None})
 
     def addAgent(self, agent):
         self.agents.append(agent)
@@ -96,6 +99,12 @@ class Sugarscape:
                 if cellMaxCapacity > self.environment.findCell(i, j).maxSugar:
                     self.environment.findCell(i, j).maxSugar = cellMaxCapacity
                     self.environment.findCell(i, j).sugar = cellMaxCapacity
+
+    def checkActiveDiseases(self):
+        for agent in self.agents:
+            if len(agent.diseases) > 0:
+                return True
+        return False
 
     def configureAgents(self, numAgents):
         if self.environment == None:
@@ -214,11 +223,24 @@ class Sugarscape:
         self.environment.findCellNeighbors()
         self.environment.findCellRanges()
 
+    def countInfectedAgents(self, disease):
+        totalInfected = 0
+        for agent in self.agents:
+            for agentDisease in agent.diseases:
+                if disease == agentDisease["disease"]:
+                    totalInfected += 1
+        return totalInfected
+
     def doTimestep(self):
         self.removeDeadAgents()
         self.replaceDeadAgents()
         if self.timestep >= self.maxTimestep:
             self.toggleEnd()
+            return
+        if self.checkActiveDiseases() == True:
+            for disease in self.diseases:
+                disease.resetRStats()
+        else:
             return
         if "all" in self.debug or "sugarscape" in self.debug:
             print(f"Timestep: {self.timestep}\nLiving Agents: {len(self.agents)}")
@@ -254,6 +276,7 @@ class Sugarscape:
         self.runtimeStats["environmentWealthCreated"] = environmentWealthCreated
         self.runtimeStats["environmentWealthTotal"] = environmentWealthTotal
         logString = '\t' + json.dumps(self.runtimeStats) + "\n]"
+        self.diseasesStats.sort(key=lambda d: d["disease"])
         if self.logFormat == "csv":
             logString = ""
             # Ensure consistent ordering for CSV format
@@ -263,16 +286,24 @@ class Sugarscape:
                 else:
                     logString += f",{self.runtimeStats[stat]}"
             logString += "\n"
+            # Entire disease log written at the end of the simulation to sort it by disease instead of timestep
+            for disease in self.diseasesStats:
+                diseaselogString = ""
+                for stat in disease.values():
+                    if diseaselogString == "":
+                        diseaselogString += f"{stat}"
+                    else:
+                        diseaselogString += f",{stat}"
+                diseaselogString += "\n"
+                self.diseaselog.write(diseaselogString)
+        else:
+            groupedDiseases = self.groupDiseases()
+            self.diseaselog.write(self.formatDiseaseJSON(groupedDiseases))
         self.log.write(logString)
         self.log.flush()
         self.log.close()
-
-    def endSimulation(self):
-        self.removeDeadAgents()
-        self.endLog()
-        if "all" in self.debug or "sugarscape" in self.debug:
-            print(str(self))
-        exit(0)
+        self.diseaselog.flush()
+        self.diseaselog.close()
 
     def findActiveQuadrants(self):
         quadrants = self.configuration["environmentStartingQuadrants"]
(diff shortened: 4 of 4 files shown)


amuller26:
I ran into an issue where if you change the number of starting diseases per agent, in sugarscape.py, each disease is added multiple times. The number of times it's added is not correlated with the number of starting diseases per agent either.

colinhanrahan:
The intended behavior is that `startingDiseases` is the master pool of diseases and `startingDiseasesPerAgent` is how many diseases each agent will be endowed with at the beginning of the simulation. So if there are 25 diseases and 10 diseases per agent, each agent should catch a random 10 of the 25 diseases (and they will not catch diseases they are immune to). Can you send me a config where you're seeing this behavior?

amuller26:
@colinhanrahan Everything in config.json is the same except startingDiseasesPerAgent is anything but [0, 0].

colinhanrahan:
Everything I'm seeing with `"startingDiseasesPerAgent" != [0, 0]` matches up with pg. 147 in the book. Can you elaborate a little more on what's happening?

amuller26:
sugarscape.py, line 178-201
`diseases = []
        for i in range(numDiseases):
            diseaseID = self.generateDiseaseID()
            diseaseConfiguration = diseaseEndowments[i]
            newDisease = disease.Disease(diseaseID, diseaseConfiguration)
            diseases.append(newDisease)
        startingDiseases = self.configuration["startingDiseasesPerAgent"]
        minStartingDiseases = startingDiseases[0]
        maxStartingDiseases = startingDiseases[1]
        currStartingDiseases = minStartingDiseases
        for agent in self.agents:
            random.shuffle(diseases)
            for newDisease in diseases:
                if len(agent.diseases) >= currStartingDiseases and startingDiseases != [0, 0]:
                    currStartingDiseases += 1
                    break
                hammingDistance = agent.findNearestHammingDistanceInDisease(newDisease)["distance"]
                if hammingDistance == 0:
                    continue
                agent.catchDisease(newDisease)
                self.diseases.append(newDisease)
                if startingDiseases == [0, 0]:
                    diseases.remove(newDisease)
                        break`
The disease loop is in the agent loop, so if an agent gets assigned the same disease as another agent, a duplicate is appended to self.diseases

amuller26:
@colinhanrahan sugarscape.py, line 178-201

> diseases = []
        for i in range(numDiseases):
            diseaseID = self.generateDiseaseID()
            diseaseConfiguration = diseaseEndowments[i]
            newDisease = disease.Disease(diseaseID, diseaseConfiguration)
            diseases.append(newDisease)
        startingDiseases = self.configuration["startingDiseasesPerAgent"]
        minStartingDiseases = startingDiseases[0]
        maxStartingDiseases = startingDiseases[1]
        currStartingDiseases = minStartingDiseases
        for agent in self.agents:
            random.shuffle(diseases)
            for newDisease in diseases:
                if len(agent.diseases) >= currStartingDiseases and startingDiseases != [0, 0]:
                    currStartingDiseases += 1
                    break
                hammingDistance = agent.findNearestHammingDistanceInDisease(newDisease)["distance"]
                if hammingDistance == 0:
                    continue
                agent.catchDisease(newDisease)
                self.diseases.append(newDisease)
                if startingDiseases == [0, 0]:
                    diseases.remove(newDisease)
                        break

The disease loop is in the agent loop, so if an agent gets assigned the same disease as another agent, a duplicate is appended to self.diseases

amuller26:
I'm not sure if this is an error, but when `configuration['startingDiseasesPerAgent'] != [0,0]`, it isn't guaranteed that all diseases will get used. Let's say `configuration["startingDiseases"] = 50`, `self.diseases` might only use 45. Is that normal or to be expected?

colinhanrahan:
It's not an error with the current implementation, but we could change the implementation if necessary. We endow each agent with `startingDiseasesPerAgent` random diseases that they are not immune to from the master list of diseases, so there are two causes for unused diseases:
- all agents are immune to the disease, so they do not contract it
- some or all agents are not immune to the disease, but the disease is never selected by chance

amuller26:
I consolidated the diseases' metrics so far into the logfile. The naming scheme so far is runtimeStats["disease{disease.ID}{metric}"].
The only metric so far is the R-Value, waiting on Dr. R's response for which other metrics to track. 
With the current naming scheme, in the JSON format, it will show up disease0, disease1, disease2, disease3, etc. However, with the CSV format, it lists them disease0, disease1, disease10, disease11, etc. Is there a better way to name the disease metrics?

colinhanrahan:
Yeah, it looks like CSV variables are sorted alphabetically (`agentAgingDeaths,agentCombatDeaths,agentDiseaseDeaths`) while JSON variables are intentionally ordered (`"timestep": 0, "population": 250, "meanMetabolism": 2.49`). There are two approaches you could take:
1. Add the disease stats _after_ the runtime stats are sorted. Disease stats are added on lines 64-65 of `__init__`:
```python
diseaseStats = {f"disease{disease.ID}RValue": 0 for disease in self.diseases}
self.runtimeStats.update(diseaseStats)
```
And runtime stats are sorted alphabetically in `endLog`. You could just add the disease stats after the sort in `endLog` instead of in `__init__`.

2. When naming disease variables, pack the disease IDs with 0s on the left to maintain alphabetical order. So if there were 11-100 diseases, you would add one extra 0 to diseases 0-9: `disease00, disease01, disease02...`. This should keep them in order.

+3. Maybe the runtime stats don't need to be sorted at all — you could ask NKH when he becomes available again. They should be in a consistent order regardless so I don't understand the comment on line

amuller26:
I started implemented my feature to start diseases at different set timesteps. However, if the diseaseStartTimeframe, it either won't initialize the rest of the diseases (if diseaseStartTimeframe = [0, 1] for example]. Or, if it starts at any timestep past 0, it will only initialize half of the diseases it's supposed to (diseaseStartTimeframe = [1, 1] for example). I am not sure where I am going wrong. It's reading the infectAgents() function and looks like it's doing everything right, but maybe I've been staring at it too long. Some suggestions and advice would be great.
I have everything logged to a separate file so I can look at the numbers and make sure it's correct. Once I get this sorted I will add all of the data to the main log.

amuller26:
I'm also not sure where to put the `infectAgents()` function because it doesn't appear visible until the day after the disease is initialized.

colinhanrahan:
How is the start timestep per disease intented to work with `startingDiseasesPerAgent`? If `"diseaseStartTimeframe": [0, 1]`, for example, does every agent get infected with a value in the range `startingDiseasesPerAgent` on timestep 0 (selected from the pool of diseases with that starting timestep — so half the total pool), and then the same on timestep 1? Or are the setting mutually exclusive?

I'm also seeing some zombie agents that don't move but don't die — I'll investigate this, but these are always caused by agents being removed from `sugarscape.agents` before calling their own death function.

amuller26:
`"diseaseStartTimeframe": [0, 1]` works similarly to the pollutionTimeframe. The disease is assigned a timestep to be initialized and infect however many agents. Agents will still be infected with however many diseases depending on `startingDiseasesPerAgent` but they might not all be at the same timestep. If `"startingDiseasePerAgent: [0, 5]` and `"diseaseStartTimeframe": [0, 5]`, then an agent could be infected by up to 5 diseases within timesteps 0-5. 

The disease is assigned a randomized timestep and then entered in a 2D array called `diseasesCount` where `infectAgents()` will use only `diseasesCount[timestep`.

colinhanrahan:
`self.infectAgents` should be moved to the top of `self.doTimestep` — stepping forward on the GUI is not part of the `runSimulation` loop and calls `sugarscape.doTimestep` directly.

amuller26:
I moved it to after `self.doTimestep` and something's still wrong. It looks like the diseases were initialized on the correct timestep but the GUI didn't recognize them and the stats are logged the timestep after. It also looks like the timestep after is day 2 of the disease in the simulation.

amuller26:
Now that it's written to the correct timestep, there's an issue where if a disease is initialized on any timestep except 0, a couple of them will duplicate, and I have no clue how.

colinhanrahan:
I saw that you added another comment through my notifications, but I can't see it right now. If you deleted it please disregard.

The diseases are introduced at the beginning of the timestep. If you're checking the diseases per agent at the end of the timestep (or after any agents have completed their their timestep), it's likely that the diseases will have spread ("duplicated").

If `"startingDiseasesPerAgent" != [0, 0]`, then new diseases can infect multiple agents at the same time. Otherwise, they should only infect one agent when they are introduced into the population.

Do either of those fix your issue? If not, can you give me some more detail?

amuller26:
There are some agents who do not move, and I don't know if it's because they can't move or they're zombies.

colinhanrahan:
Zombie agents will have negative sugar and metabolism and their age won't increase over time. Sick agents with 0 functional range will have 0 vision and/or 0 movement.

colinhanrahan:
Add `diseaselog.csv` and `diseaselog.json` to the `make clean` command:
```bash
LOGS = log.csv log.json diseaselog.csv diseaselog.json
```

amuller26:
I modified agent behavior to avoid any cell with sick agents in the vicinity when they're not sick, and to avoid only areas with the same tribe if the agent is sick. @colinhanrahan Can you please check that it is implemented correctly?

colinhanrahan:
Screenshot 2024-08-23 at 5 53 27 PM

Running `disease_basic.json` (after fixing the breaks on my two active PRs), agents refuse to move next to any agent with a disease and end up in a gridlock where many can't move at all. This seems like a natural consequence of the logic you introduced rather than a bug. Thoughts?

nkremerh:
A thought for the self-quarantining and the exiling behavior:

These should be configurable parameters for the agent (such as `self.quarantineFactor`) which impact the scores for potential cells. A considerate sick agent (i.e. one with a `quarantineFactor` of 1) will impose strong score penalties to cells next to others. Likewise, a healthy agent with a high `quarantineFactor` will impose strong score penalties to cells next to sick agents.

That way, there's less likely to be gridlock. Agents may not find too many cells with good scores, but they're still likely to move since those cells next to sick/healthy individuals aren't completely removed from consideration. They're simply less appealing.

nkremerh:
@amuller26 I made some changes to the code but otherwise performed a nasty rebase of this pull request atop the master branch. Please double check for functionality and correctness before I continue reviewing this PR.

amuller26:
@nkremerh Everything looks accurate with my original branch. The only thing missing was updating the runtime stats, which I fixed.

nkremerh:
Let's get this merged soon. Please resolve the merge conflicts and let me know when you're ready for me to finalize this PR. I'll generalize some of the mechanics you introduced.

nkremerh:
@amuller26 I went through the pull request and made a generalization and style pass on the mechanisms introduced here. I'm fairly certain I kept everything except the per-disease metric tracking. There's now a way of tracking a disease as an experimental group (by using ``disease1`` in the ``experimentalGroup`` property, for instance). All other diseases have their stats clumped together to form the control group. That resolves the ballooning log size problem.

I also left behind categorizing the diseases into susceptible, infected, recovered, and immune buckets. It is clear to me that the SIR model is driving the disease features with this PR, but I do not yet see an interesting use case for logging all that extra info explicitly in the agent data structure (and spending the time to do so). If this is a necessary component, please let me know. I can think it over and work it in.

I have a copy of your original commits saved locally, but I would also recommend you do the same before pulling this reworked branch. I translated the features in a fresh branch, so your previous history is wiped out here.

PR #119, opened 2024-08-21
Adds option to specify experimental group under study

nkremerh:
Adds separate logging of experimental and control groups when specified
Modifies naming of sick agent log entry to differentiate it from grand total entries
Adds consideration of basic income when calculating agent time to live

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -840,6 +840,13 @@ class Agent:
         # If no sugar or spice metabolism, set days to death for that resource to seemingly infinite
         sugarTimeToLive = self.sugar / sugarMetabolism if sugarMetabolism > 0 else sys.maxsize
         spiceTimeToLive = self.spice / spiceMetabolism if spiceMetabolism > 0 else sys.maxsize
+        # If an agent has basic income, include the income for at least as many timesteps as they can already survive
+        if self.universalSugar != 0:
+            sugarIncome = (sugarTimeToLive * self.universalSugar) / self.cell.environment.universalSugarIncomeInterval
+            sugarTimeToLive = (self.sugar + sugarIncome) / sugarMetabolism if sugarMetabolism > 0 else sys.maxsize
+        if self.universalSpice != 0:
+            spiceIncome = (spiceTimeToLive * self.universalSpice) / self.cell.environment.universalSpiceIncomeInterval
+            spiceTimeToLive = (self.spice + spiceIncome) / spiceMetabolism if spiceMetabolism > 0 else sys.maxsize
         timeToLive = min(sugarTimeToLive, spiceTimeToLive)
         if ageLimited == True:
             timeToLive = min(timeToLive, self.maxAge - self.age)
@@ -1093,6 +1100,21 @@ class Agent:
         if self.sugar >= self.startingSugar and self.spice >= self.startingSpice and self.age >= self.fertilityAge and self.age < self.infertilityAge and (self.fertilityFactor + self.fertilityFactorModifier) > 0:
             return True
         return False
+
+    def isInGroup(self, group, notInGroup=False):
+        membership = False
+        if group == "depressed":
+            membership = self.depressed
+        elif group == "female":
+            membership = True if self.sex == "female" else False
+        elif group == "male":
+            membership = True if self.sex == "male" else False
+        elif group == "sick":
+            membership = self.isSick()
+
+        if notInGroup == True:
+            membership = not membership
+        return membership
 
     def isLender(self):
         # If not a lender, skip lending
--- a/sugarscape.py
+++ b/sugarscape.py
@@ -62,11 +62,12 @@ class Sugarscape:
                              "meanSocialHappiness": 0, "meanFamilyHappiness": 0, "meanConflictHappiness": 0, "meanAgeAtDeath": 0, "seed": self.seed, "totalWealthLost": 0,
                              "totalMetabolismCost": 0, "agentsReplaced": 0, "agentsBorn": 0, "agentStarvationDeaths": 0, "agentDiseaseDeaths": 0, "environmentWealthCreated": 0,
                              "agentWealthTotal": 0, "environmentWealthTotal": 0, "agentWealthCollected": 0, "agentWealthBurnRate": 0, "agentMeanTimeToLive": 0,
-                             "agentTotalMetabolism": 0, "agentCombatDeaths": 0, "agentAgingDeaths": 0, "totalSickAgents": 0}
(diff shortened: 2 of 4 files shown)


nkremerh:
@colinhanrahan @asipes33 @amuller26 Please review.

Colin, check for efficiency gains (which are still stylistically compliant).
Abi, check for correctness for the depression use case.
Anna, check for correctness for the epidemiology use case.

agent.py
amuller26, line 1: I added a boolean `self.diseaseDeath = False` to keep track if the agent died while infected, as per our last meeting with Dr. Rutherford. Then I switched it to true in `doDeath()` if the agent died while sick.

sugarscape.py
```diff
@@ -893,50 +911,41 @@ def updateRuntimeStats(self):
             agentDiseaseDeaths += 1 if agent.causeOfDeath == "disease" else 0
```
amuller26, line 911: Modify this to `agentDiseaseDeaths += 1 if agent.diseaseDeath == True else 0` since there is no way to detect if an agent dies from disease.

sugarscape.py
```diff
@@ -893,50 +899,48 @@ def updateRuntimeStats(self):
...
+            groupStats = {}
+            for key in runtimeStats.keys():
+                groupKey = groupString + key[0].upper() + key[1:]
+                groupStats[groupKey] = runtimeStats[key]
+            runtimeStats = groupStats
+
+        for key in runtimeStats.keys():
+            self.runtimeStats[key] = runtimeStats[key]
+
+    def updateStats(self):
```
colinhanrahan, line 938: Consider renaming `updateStats` and/or `updateRuntimeStats` to better differentiate between them if you can come up with something more descriptive — they're both runtime stats. Maybe `updateRuntimeStats` could include something group-related.

sugarscape.py
```diff
@@ -1103,6 +1107,14 @@ def verifyConfiguration(configuration):
         removeDecisionModels = uniqueAgentDecisionModels[maxColors:]
         configuration["agentDecisionModels"] = [i for i in configuration["agentDecisionModels"] if i not in removeDecisionModels]
 
+    # Ensure experimental group is properly defined or otherwise ignored
+    if configuration["experimentalGroup"] == "":
+        configuration["experimentalGroup"] = None
+    elif configuration["experimentalGroup"] not in ["depressed", "female", "male", "sick"]:
```
colinhanrahan, line 1113: `None` should be included in the values to check. If `experimentalGroup` is left out of the config (e.g. in examples), debug will print `Cannot provide separate log stats for experimental group None. Disabling separate log stats.` The `None` value was correctly provided by `sugarscape.py` — this is intended behavior and not a user error.

colinhanrahan (commented):
Looks good. Efficiency gains are more likely to be found elsewhere in the code — in my testing, `isInGroup` takes up about 0.13% of the runtime and `updateStats` takes up about 4%.

sugarscape.py
```diff
@@ -893,50 +899,48 @@ def updateRuntimeStats(self):
...
+        runtimeStats = {"agentAgingDeaths": agentAgingDeaths, "agentCombatDeaths": agentCombatDeaths, "agentDiseaseDeaths": agentDiseaseDeaths, "agentMeanTimeToLive": agentMeanTimeToLive,
+                        "agentStarvationDeaths": agentStarvationDeaths, "agentTotalMetabolism": agentTotalMetabolism, "agentWealthBurnRate": agentWealthBurnRate,
+                        "agentWealthCollected": agentWealthCollected, "agentWealthTotal": agentWealthTotal, "maxWealth": maxWealth, "meanAge": meanAge, "meanAgeAtDeath": meanAgeAtDeath,
+                        "meanConflictHappiness": meanConflictHappiness, "meanFamilyHappiness": meanFamilyHappiness, "meanHappiness": meanHappiness, "meanHealthHappiness": meanHealthHappiness,
+                        "meanMetabolism": meanMetabolism, "meanMovement": meanMovement, "meanSocialHappiness": meanSocialHappiness, "meanTradePrice": meanTradePrice, "meanWealth": meanWealth,
+                        "meanWealthHappiness": meanWealthHappiness, "meanVision": meanVision, "minWealth": minWealth, "population": numAgents, "sickAgents": sickAgents, "tradeVolume": tradeVolume
+                        }
```
asipes33, line 912: Do we get a stat for how many agents died in total for a given timestep? Not just the breakdown of the cause of death? I need a total death count for one of the graphs.

asipes33:
Other than my one comment, everything looks correct for my data.

amuller26:
Everything looks good on the epidemiology side.

sugarscape.py
nkremerh: Leaving a note here to include this in the enhanced disease mechanics in your PR. This merge is going to cause conflicts with that since there are so many (good) changes in your PR.

sugarscape.py
```diff
@@ -893,50 +899,48 @@ def updateRuntimeStats(self):
...
+
+        if group == None:
+            self.runtimeStats["agentsBorn"] = self.agentsBorn
+            self.runtimeStats["agentsReplaced"] = self.agentsReplaced
+            self.runtimeStats["environmentWealthCreated"] = environmentWealthCreated
+            self.runtimeStats["environmentWealthTotal"] = environmentWealthTotal
+            self.runtimeStats["giniCoefficient"] = self.updateGiniCoefficient()
+            self.runtimeStats["timestep"] = self.timestep
+            self.runtimeStats["totalMetabolismCost"] += totalMetabolismCost
+            self.runtimeStats["totalWealthLost"] += totalWealthLost
```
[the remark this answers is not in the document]
nkremerh: Could be useful to do a per-group Gini calculation. I'll leave it out of this PR since it's out of scope for the immediate need (and looks like a bit of a gnarly change).

The two running total metrics have been removed. They were never used. Good eye.

sugarscape.py
nkremerh: Done.

sugarscape.py
nkremerh: Done.

sugarscape.py
nkremerh: Done.

nkremerh:
@colinhanrahan @asipes33 One last look, please.

colinhanrahan (approved):
Everything looks good.

colinhanrahan:
To me, `agentsReplaced`, `femaleAgentsReplaced` etc. reads as "agents of this group who died and had to be replaced, but the metric is actually tracking "replacement agents that are of this group." Maybe `replacementAgents` would be a better name?

nkremerh:
There are no sick agents born, you're right. This is a generalized solution, so there will be certain metrics which are useless for analysis after the fact.

PR #125, opened 2025-03-02
Adds generalized conditions to merge depression and disease mechanisms

nkremerh:


nkremerh:
@goolabjamun @asipes33 This could use a review.

nkremerh:
@colinhanrahan Does anything seem broken here?

colinhanrahan:
- Should `sugarscape.diseases` be renamed to `conditions` since it includes depression?
- I interpreted `"agentDepressionPercentage"` to mean that 30% is a value of `30.0`, but it's actually `0.30`, which is more of a proportion. If we're keeping this name, consider adding a note in the README to clarify that the percentage is between `0.0` and `1.0`.

Everything works as far as I can tell. If more specialized conditions like depression are being added in the future, it would be useful to add modularity for those conditions rather than hardcoding depression as an agent variable / endowment.

nkremerh:
I made the depression percentage clearer in the README.

The `sugarscape.diseases` name will remain as the main purpose of it is to hold diseases. Major depressive disorder can be classified as a disease (just not an infectious kind). Other conditions like losing a limb (movement reduction) or going blind (vision reduction) could be possible but not be diseases. Those are future ideas, so no rush to change things now to accommodate.

PR #127, opened 2025-03-05
Fixes resizing logic to only resize after user releases window

colinhanrahan:
While we're updating resizing logic, can we also check that `doCrossPlatformWindowSizing` is still necessary on some operating systems? The GUI works fine when I remove the function on macOS.

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -9,6 +9,8 @@ class GUI:
         self.canvas = None
         self.grid = [[None for j in range(self.sugarscape.environmentHeight)]for i in range(self.sugarscape.environmentWidth)]
         self.window = None
+        self.doubleClick = False
+        self.resizeID = None
 
         sugarAndSpiceColors = self.findSugarAndSpiceColors("#F2FA00", "#9B4722")
         pollutionColors = self.findColorRange("#FFFFFF", "#803280", 0, 20)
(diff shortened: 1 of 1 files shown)


gui.py
```diff
@@ -385,6 +387,15 @@ def doPlayButton(self, *args):
...
         self.doTimestep()
 
+    def doResize(self, event):
+        # Do not resize if capturing a user input event but the event does not come from the GUI window
+        if event != None and (event.widget != self.window or (self.screenHeight == event.height and self.screenWidth == event.width)):
+            return
+
+        if self.resizeID != None:
+            self.window.after_cancel(self.resizeID)
+        self.resizeID = self.window.after(20, self.resizeInterface)
```
colinhanrahan, line 397: The delay is currently set to 20ms, but it should be higher than the highest practical polling interval across all supported devices. If that number is higher, we should increase the delay.

Although, if you're fine with the window occasionally unecessarily resizing when the polling interval is very high, we can set the delay to a lower value.

nkremerh:
If the cross-platform checking isn't breaking anything, I see no reason to delete it.

colinhanrahan:
@nkremerh A small nitpick — the `after` parameter is a polling interval (in ms) rather than a polling rate (in Hz).

nkremerh:
I will push a commit renaming the variable. 10ms seemed to look better across two machines for me, but I can bump it back up if desired.

colinhanrahan:
10ms is good!

PR #128, opened 2025-04-11
Interactions

asipes33:


nkremerh:
@asipes33 I handled the divergent history conflict and generalized the mechanism (so we can compare depressed to non-depressed agents, diseased to non-diseased, etc.).

Please let me know if this accidentally drops any desired functionality.

nkremerh:
@goolabjamun Please also double-check when you have time.

